PR #442, opened 2015-07-28
New portuguese translations to newly added terms

danimaribeiro:
I have more terms I've translated, but I wait too long, now has some conflicts.
From now on I'm going to send more small PR with the translations, is it ok?


burisu:
Yes, it's better with small PR. Thank you again.

PR #554, opened 2016-04-03
User sign up confirmed by admin

jonathanpa:
* Miss only to find how to change label of Update button in `field_set` by Approved.


burisu:
@jonathanpa How the administrator check and validate the registration. When I tried it, no mails send, and users account have been created. I can see a button to validate users.

jonathanpa:
@burisu For the moment, "all" the users invited and signed up but not yet confirmed are listed in the users list `backend/users`.

Like invitees, unconfirmed users are instance of `User` model but with `!signup_at.present? == true`.

To list the unconfirmed users: `backend/registrations`, then Edit and click on Update (want to change its label to Approve but didn't find how to do, see my question about it at the top of the PR comments).

You are right, by being extremely focused to restrict unconfirmed users to access, I forgot to add an email notification for new sign up to admins.

PR #745, opened 2016-06-15
Adds product nature variant components

Guichevrat:
Adding the possibility to add a trouble and its description during an intervention. Showing it in the intervention's list. 
Creating `product_nature_variant_component` table to add components into another.
Adding the views to show a product's composition.


burisu:
@Guichevrat, you need to remove `20160607091145_add_lifespan_to_product_nature_variants` migration, before I can merge anything. I can't find your nomenclature migration for missing indicators. Maybe you can do all at same time.

burisu:
Now in #851 .

PR #992, opened 2016-08-05
Migrate to Rails 5

burisu:
Time to go on


Aquaj:
One thing we'll have to be careful of is that in Rails 5 belongs_to associations are required by default in validations.
I haven't checked in details but I think it might cause us trouble on polymorphic tables with belongs_to. (If the table Model is present as ModelA and ModelB and ModelA has a belongs_to but ModelB doesn't.)

How to opt-out where needed : http://blog.bigbinary.com/2016/02/15/rails-5-makes-belong-to-association-required-by-default.html#opting-out-of-this-default-behavior-in-rails-5

burisu:
See #2118 from now.

PR #1006, opened 2016-09-06
Add a low-level log system for API requests and responses

Aquaj:
To track all exchanges and known exactly what has worked and what has not, we'll tack everything.

```ruby
class CallMessage < ActiveRecord::Base
  belongs_to :request, class_name: 'Call'
  # attrs: body, headers, status, sender_ip, treated_at, sent_at (?), ...
end
```
In API, a concern dedicated to automate logging of the messages must be created in order to be used in `Api` namespace and other extensions in plugins.
```ruby
class MyApiController < ActionController::Base
  include CallLoggable
  ...
end
```


Aquaj:
Will be fixed by #1007.

PR #1146, opened 2016-10-02
Add missing install command for deps installation.

fxprunayre:


fxprunayre:
Let me know if I should better target another branch eg. master ? (I'm new to the project and not sure about the merging strategy you have).

PR #1157, opened 2016-10-04
Sepa transfer for purchases

jonathanpa:
Some notes:
* A `t.column :sepa` in the `BackendOutgoingPaymentModesController list` raises an `undefined method `text' for "<div class=\"checkbox-true\"></div>":ActiveSupport::SafeBuffer`
* For the dropdown "Pay" button in the `OutgoingPaymentList` index view, I didn't way how to make
the button dropdownable with the existing helper. Consequently, I take the first `OutgoingPaymentMode` in SEPA.
  * Should I take the `active == true` `OutgoingPaymentMode` ? I see in fixtures they have all `active: false`.
  * I take the purchases whose `state` is `order`. When should we move the `state` to `invoice` ?
* For printing, I tried to follow what was done for the purchases or the inventories without success. By the way, to print purchase or inventory raises me an error on my machine.


app/controllers/backend/outgoing_payment_modes_controller.rb
```diff
@@ -31,6 +31,7 @@ class OutgoingPaymentModesController < Backend::BaseController
       t.column :name
       t.column :cash, url: true
       t.column :with_accounting
+      # t.column :sepa
```
jonathanpa: If enable I get an `undefined method 'text' for "<div class=\"checkbox-true\"></div>":ActiveSupport::SafeBuffer`.
Is `t.column :sepa` the correct way to add a "boolean" column to the table ?

app/controllers/backend/outgoing_payment_modes_controller.rb
```diff
@@ -31,6 +31,7 @@ class OutgoingPaymentModesController < Backend::BaseController
       t.column :name
       t.column :cash, url: true
       t.column :with_accounting
+      t.column :sepa
```
Aquaj: Appears to be working. (@ 56eebae)

app/controllers/backend/outgoing_payment_modes_controller.rb
Aquaj: See previous review on same line. Fixed as of now (56eebae), probably by 3da1d44.

Aquaj:
Reviewed. Would like another review (cc @igkyab ).

igkyab:
Reviewed too. I merge the branch.

PR #1751, opened 2017-03-24
Update tax declaration computation

lcoq:


lcoq:
Rebased version on #1754

PR #1752, opened 2017-03-24
Added navigation for entities.

DucMirack:


app/views/backend/entities/show.html.haml
```diff
@@ -1,3 +1,8 @@
+- if resource.nature== 'contact'
+  - navigation resource, order: :full_name, naming_method: :label, scope: :contacts
+- elsif resource.nature== 'organization'
```
Aquaj: `==` should be spaced around.

`resource.nature == 'contact'` instead of current code would be better.

app/views/backend/entities/show.html.haml
```diff
@@ -1,3 +1,8 @@
+- if resource.nature== 'contact'
```
Aquaj: `==` should be spaced around.

`resource.nature == 'contact'` instead of current code would be better.

Aquaj:
You use `if X == A do something_with A` + if `X == B do something_with B` why not just `something_with X`?

Aquaj:
Refacto'd as it was a quick change.

Will merge during release process.

Issue #1771, opened 2017-04-04
interventions#index (ActionView::Template::Error) "undefined method `parameters' for nil:NilClass"


igkyab:
## Error
An ActionView::Template::Error occurred in interventions#index:

  undefined method `parameters' for nil:NilClass
  app/models/intervention.rb:452:in `completely_filled?'

---
## Environment
- **Version :** 2.39.1
- **Server :** production
- **Date :** 03/04/2017

burisu:
@thib44 , can you explain how you find the solution in #1774?

PR #1774, opened 2017-04-04
Add a migration to rename old animal_changing interventions

thib44:
https://github.com/ekylibre/ekylibre/issues/1771


thib44:
@burisu The procedure `animal_changing` was no longer referenced, and the good one was `animal_group_changing`.

Issue #1757, opened 2017-03-31
interventions#create (NoMethodError) "undefined method `id' for nil:NilClass"


igkyab:
## Error
A NoMethodError occurred in interventions#create:

  undefined method `id' for nil:NilClass
  app/models/journal_entry_item.rb:242:in `attributes_for'

---
## Environment
- **Version :** 2.39.1
- **Server :** production
- **Date :** 31/03/2017

PR #1776, opened 2017-04-05
Account = nil : Add a condition if variant is not storable + debug sort_by in journal_entries/_form

thib44:
See #1757 issue


thib44:
https://github.com/ekylibre/ekylibre/issues/1757

PR #1781, opened 2017-04-07
Loan repayment : translation ok (in edit), error if no started_on ok, validation of amount in loan

thib44:


thib44:
@burisu  I change the `name` to `due_on` because the attribute `name` don't exist for `LoanRepayment`.
Ok for the test, and I'm going to find a solution for the `try`.

thib44:
@burisu About the `try`, in this case to avoid error message, I have two solutions : 
`self.ongoing_at ||= started_on.try(:to_time)`, or : `self.ongoing_at ||= started_on.to_time if started_on`, wich one do you prefer ? Why ?

PR #1853, opened 2017-05-03
Update of accountancy close after review

Aquaj:
Issues noted:
- [x] Carry-forward entries are on last day of closed year instead of first day of new year.
- [x] Cash graph on Accountancy dashboard is messed up by closing the year. (appeared to be due to the first point)
- [x] Accountancy fog validation should be asynchronous. (currently under development by @thib44)
- [x] "Hide lettered elements" is currently hiding partially lettered ones too.

### DIFF
--- a/app/assets/javascripts/backend/loan_date.js.coffee
+++ b/app/assets/javascripts/backend/loan_date.js.coffee
@@ -1,0 +1,11 @@
+((E, $) ->
+  'use strict'
+
+  $(document).ready ->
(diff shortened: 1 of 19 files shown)


Aquaj:
Closed 3rd point because @thib44 opened Pull Request #1786.

Aquaj:
Removed 3rd point ("Closing a financial year should be asynchronous.") as it will be in its own PR.

PR #1857, opened 2017-05-04
Feature/add sum in lists

thib44:
Add sum with active list in :
- `general-ledger` 
- `accounts`
- `incoming_payments`
- `outgoing_payments`


thib44:
TODO: 
`interventions`

PR #1875, opened 2017-05-11
Creating a debit + a credit entries instead of a single summed entry for non-letterable accounts in FinancialYear#close

Aquaj:
See Taiga#507.


burisu:
Merged in #1886.

PR #1876, opened 2017-05-11
Better handling of partial lettering in FinancialYear#close.

Aquaj:
- Fixed bug where the affair would be updated but wouldn't update its entry_items. (s/update_column/update)
- Fixed bug where all non-lettered carried-forward entry items would be lettered A*.


burisu:
Merged in #1886.

PR #1877, opened 2017-05-11
Added consideration for the lettered/able accounts during FinancialYear#close process.

Aquaj:
This is in accordance to Step 2-b of @AntoineGrz 's spec and as written in Taiga#506.

This has a horrific impact on performance since we have a lot more entries to compute for and then write. As such this shouldn't be deployed to production before #1855 is.


Aquaj:
FinancialYear#close on our usual big dataset finished in 5511614ms (ActiveRecord: 2729448.9ms).
So about 91mns, 45 of which are ActiveRecord.

The 45 minutes of non-ActiveRecord code execution can probably be reduced greatly by extracting the process into a ServiceObject, which would allow for caching in instance variables and would also de-complexify current code.

burisu:
Merged in #1886.

PR #1944, opened 2017-06-10
Adds permanent list sum during scrolling

burisu:


burisu:
Effective merge of #1879 in master

Issue #1556, opened 2017-01-06
Crops killed in Ekylibre are sent in Zero plant countings


PH64:
Even if plants have a dead date in Ekylibre, dead plants are are in Zero plant counting

### 1- How to reproduce?
In Zero application, in plant counting, dead crops appears while they shouldn't. The API (`api/v1/plants`) should filter records using current datetime over the range `born_at..dead_at`.

### 2- What is really expected?
Dead crops should not appear in `GET api/v1/plants` action and so, in crop fields in Zero application. 

### 3- How to test it?
Check in Zero app, dead crops should not appear in crop selection in crop counting function.

PR #1969, opened 2017-06-17
Issue fixed - Crops killed in Ekylibre are sent in Zero plant countin…

vivek2007:
…gs #1556


Aquaj:
PR is not relevant anymore as issue as been reassigned.

PR #1970, opened 2017-06-19
Disable bookkeeping for purchases which are used in tax declarations

thib44:


app/models/purchase.rb
```diff
@@ -239,6 +239,13 @@ def self.affair_class
     "#{name}Affair".constantize
   end
 
+  def bookkeepable?
+    with_accounting &&
+    invoice? &&
+    items.any? &&
+    TaxDeclarationItemPart.where(journal_entry_item_id: journal_entries.last.items.map(&:id)).any?
```
burisu: Prefer `.select` instead of `.map` here because Arel will build only one query, that's the best!

app/models/purchase.rb
```diff
@@ -240,6 +240,13 @@ def self.affair_class
     "#{name}Affair".constantize
   end
 
+  def bookkeepable?
+    with_accounting &&
+    invoice? &&
+    items.any? &&
+    TaxDeclarationItemPart.where(journal_entry_item_id: journal_entries.last.items.select(&:id)).any?
```
Aquaj: This should be `select(:id)` instead of `select(&:id)`.

PR #1971, opened 2017-06-20
Clear code translation missing

vivek2007:


Aquaj:
Obsolete because of #1975.

PR #1981, opened 2017-06-21
Prefill fields in PurchasePayments when generated through a Purchase

### LINKED ISSUE #1943: Some fields in outgoing payments are no longer pre-filled.
When you add a new purchase-payment from a purchase, some fields are no longer pre-filled : 
the 2 dates and the manager.

Aquaj:
Fixes #1943.
- Added a way to pass prefill options to the links in the new_x buttons in `affair_of` helper.
- Prefilled `responsible`, `paid_at` and `to_bank_at` in new PurchasePayment.


EmmaDorie:
Please, could someone finish correcting this bug????

app/views/backend/purchases/show.html.haml
```diff
@@ -41,7 +41,7 @@
 
   - if resource.affair
     - c.cobble :affair do
-      = affair_of(resource, default: 'PurchasePayment')
+      = affair_of(resource, default: 'PurchasePayment', purchase_payment_attributes: { responsible_id: resource.responsible_id, paid_at: resource.payment_at, to_bank_at: resource.payment_at})
```
burisu: Why the option name is `<model>_attributes` and not `params` or `url_options` or something more "URLesque"

Aquaj: Do you mean something like `params: { <model>: { ... } }`? Or simply `<model>_params` or `<model>_url_options` ?
I would agree indeed for a change towards the latter but we still need a way to characterize these options by model as we might want to pass different options for each of the involved models in `affair_of`.

Maybe `url_options: { <model>: { ... }, <other_model>: { ... } }` ?

burisu: Yes, sorry, I forgot at that time that we can have other types... The most explicit name as parameter is: `new_<model>_url_options`. But your last proposal seems more sustainable.
Do you agree?

Aquaj: I agree. I will implement the last proposal.

PR #1983, opened 2017-06-21
Add `codes` column to CultivableZones, Products and Activities

Aquaj:
Adds `codes` transcoding columns to Product, CultivableZones and Activities so we can be more flexible on third-party integration.


Aquaj:
[ekylibre:feature/imepe/add-codes-to-cultivable-zones](https://trello.com/c/RFjuy6oG/178-ekylibrefeature-imepe-add-codes-to-cultivable-zones)

PR #1987, opened 2017-06-22
Progress's capabilities extension

Aquaj:
Progress now should be multi-process capable (not yet able to handle concurrent writes but there shouldn't be any because of a @read-only mechanism).

Fixed some other issues in the process, and overall made the library more robust and flexible.


Aquaj:
[ekylibre:feature/progress-no-obj-fetch](https://trello.com/c/QPvwENUT/224-ekylibrefeature-progress-no-obj-fetch)

PR #1992, opened 2017-06-28
Add management of error during the import of the first run and add progress bar

thib44:


app/mailers/support_mailers.rb
```diff
@@ -0,0 +1,14 @@
+class SupportMailers < ActionMailer::Base
+  def first_run_errors(message, time)
+    @tenant = Ekylibre::Tenant.current
+    message = message.read
+    message = message.split('; ').reject { |c| c.blank? }.last.split(', ')
+    error_time = DateTime.parse(message.first)
+    @time = time - error_time
+    @progress = message.second
+    mail(
+      to: 'support@ekylibre.com',
```
burisu: Should be configurable

app/mailers/support_mailers.rb
```diff
@@ -0,0 +1,14 @@
...
+      subject: "Erreur durant l'import du first-run"
```
burisu: No natural language in code. Use I18n.

app/mailers/support_mailers.rb
```diff
@@ -0,0 +1,14 @@
+class SupportMailers < ActionMailer::Base
```
burisu: This mailer is not a part of Ekylibre. Must be removed from here.

app/views/support_mailers/first_run_errors.html.haml
```diff
@@ -0,0 +1,12 @@
+%p
```
burisu: This files must be removed from Ekylibre

PR #2000, opened 2017-07-03
Resolved error when you create intervention show the different type of interventions. Add block in method_missing

thib44:
Added `&block` in `missing_method` in Nomen::Record so find_each can work properly again.


burisu:
Congrats for 2000th PR 👏

PR #2002, opened 2017-07-06
Fixed silent blocking error when creating multiple Fixed Assets with the same variant from a Purchase.

Aquaj:
Invoicing the parcel would fail silently if you created more than 2 assets with the same variant in the Purchase.

It created "Variant Asset" then "Variant Asset 1" then would try to create "Variant Asset 1" again for all the following assets instead of 2, 3, 4...


Aquaj:
[ekylibre:debug/fixed-asset-in-purchases](https://trello.com/c/aUPM20Vj/220-ekylibredebug-fixed-asset-in-purchases)

PR #2003, opened 2017-07-06
Update code for modify affairs from purchase view

sandy1987:
Hi, 
I have update code for modify affairs' supplier when edit the supplier from purchase view. Please check and let me know if you want any changes.


sandy1987:
Hi , 
I have move controller code into model with alone affair condition. Please check and let me know.

sandy1987:
Hello @Aquaj , 

I am little bit confuse, Let me explain the things with my understanding.....

If affair has one deal with  purchase, Then only we have to update the third into the affair. Otherwise its will change only purchase view not in the affair.

Please let me know my assumption is correct?

Aquaj:
Hello @sandy1987, I had not seen your message, sorry for that.

Yes, we only update the third of the affair if it has just one deal (the purchase that was updated). Otherwise the third changes in the purchase but does not change in the affair of the purchase.

sandy1987:
Hello @Aquaj ,

I have update the code on branch,Please check it and let me know if any thing is missing.
For Example:-
purchase_deal = affair.deals.collect(&:nature).map{|p| p  if (p.nature == "purchase")}

On the above variable we are finding all deals of that affair where deals whose nature is purchase.

Thanks

PR #2004, opened 2017-07-06
Add new dump/restore system.

Aquaj:
Reworked system to use Postgres dumps instead of YAML fixtures.

Said system should be retrocompatible with the current (v2) system thanks to Manifest files.


burisu:
[ekylibre:feature/new-dump-restore](https://trello.com/c/lXzVxFUe/242-ekylibrefeature-new-dump-restore)

PR #2007, opened 2017-07-12
Wrong Error Message When Adding new entry

sandy1987:
Hello,

I have update the code for wrong error message on the top of the new entry form like "Error!Currency can't be blank".
validate_error


sandy1987:
@Aquaj 

In this issue when we select past financial year, then its financial year not present, That the reason both validations true that the reason "Error! Currency can't be blank" Error message occur , Because without financial year we can't get the currency.

Solution => If we don't found the financial year (past financial year), So unable to get the financial year currency, But the absolute currency is present, So Please let me know can I set absolute currency to currency also.

Please confirm

Thanks

Aquaj:
Hello Sandeep,
Setting the `currency` to value of `absolute_currency` if the financial year is closed seems like an acceptable solution indeed.

sandy1987:
Hello @Aquaj ,

I have updated the code, Please check it and let me know if any other changes required.

PR #2028, opened 2017-07-24
Product nature variant valuings

thomasbirien:
For now, i create a model ProductNatureVariantValuing, and i implement this model with some information. I have begun with the model ParcelItem, i use a callback before save :calculate_average_cost_amount, with a condition (if is the first time or not). I will continue in this way.


app/views/backend/product_nature_variants/show.html.haml
```diff
@@ -24,6 +24,31 @@
         %span.value
           %i.picto{class: "picto-stock"}
           = resource.current_stock.to_s + " " + resource.unit_name
+    - if resource.current_stock == 0
+      .med-info.important
+        %span.title CUMP
```
burisu: Every human word must i18nized. You can use `AR.human_attribute_name` or `.tl` methods to localize these titles

app/views/backend/product_nature_variants/show.html.haml
```diff
@@ -24,6 +24,31 @@
...
+        %span.value 0 €
```
burisu: Why values are hard-coded? Use of localization is necessary here too. Use `.l` method to localize number or money.

app/views/backend/product_nature_variants/show.html.haml
```diff
@@ -24,6 +24,31 @@
...
+      .med-info.important
+        %span.title Valeur du stock
+        %span.value 0 €
+    - if resource.valuing
+      .med-info.important
+        %span.title CUMP
+        %span.value
+          =resource.valuing.average_cost_amount
```
burisu: Leave a space after `=` or `-` in HAML for clearness

app/views/backend/products/show.html.haml
```diff
@@ -40,6 +40,15 @@
           %span.value
             %i.picto{class: "picto-movements"}
             = link_to(resource.outgoing_parcel_item.parcel.number, controller: :parcels, action: :show, id: resource.outgoing_parcel_item.parcel.id)
+      .med-info.important
+        %span.title CUMP
```
burisu: Same thing here as in `app/views/backend/product_nature_variants/show.html.haml`.

config/routes.rb
```diff
@@ -1021,6 +1021,8 @@
 
     resources :workers, concerns: :products
 
+    resources :stocks, concerns: [:list]
```
burisu: Add resources in alphabetical order for clearness

db/migrate/20170720125756_create_product_nature_variant_valuings.rb
```diff
@@ -0,0 +1,10 @@
+class CreateProductNatureVariantValuings < ActiveRecord::Migration
+  def change
+    create_table :product_nature_variant_valuings do |t|
+      t.decimal :average_cost_amount, precision: 19, scale: 4, null: false
+      t.decimal :amount, precision: 19, scale: 4,              null: false
+      t.references :variant,                                   null: false, index: true
```
burisu: Add a foreign key constraint

db/migrate/20170720125756_create_product_nature_variant_valuings.rb
```diff
@@ -0,0 +1,10 @@
...
+      t.timestamps                                             null: false
```
burisu: Use `t.stamps` instead of `t.timestamps`. It's a homemade method to add more tracking columns.

db/migrate/20170720125756_create_product_nature_variant_valuings.rb
```diff
@@ -0,0 +1,10 @@
+class CreateProductNatureVariantValuings < ActiveRecord::Migration
+  def change
+    create_table :product_nature_variant_valuings do |t|
```
burisu: Table is here for historic, so we need to add an explicit column to track event times like: `computed_at`

db/migrate/20170801091621_add_foreign_key_to_product_nature_variant.rb
```diff
@@ -0,0 +1,5 @@
+class AddForeignKeyToProductNatureVariant < ActiveRecord::Migration
+  def change
+    add_reference :product_nature_variants, :valuing, index: true
```
burisu: A foreign key constraint must be added.
If possible, can you merge the 2 migrations in order to be consistent and update timestamp of migration in order to be last one after those of master branch?

private/test/attachments/financial_year_exchanges/11/original.csv
```diff
@@ -0,0 +1,3 @@
+Jour,Numéro de compte,Journal,Tiers,Numéro de pièce,Libellé écriture,Débit,Crédit,Lettrage
```
burisu: All these files in `private/test` must be removed. A `git checkout -f master -- private/test` should make the job

test/models/intervention_input_test.rb
```diff
@@ -58,5 +58,30 @@
 
 class InterventionInputTest < ActiveSupport::TestCase
   test_model_actions
+
+  test 'method average_cost_amount' do
+    i = {
```
burisu: Replace `i = { ... }` directly with `i_o = InterventionInput.new( ... )` and `intervention` is preferred to `i_o`.

test/models/inventory_item_test.rb
```diff
@@ -41,5 +41,11 @@
 
 class InventoryItemTest < ActiveSupport::TestCase
   test_model_actions
+
+  test 'method average_cost_amount' do
+    i = inventory_items(:inventory_items_001)
```
burisu: We hadn't discuss about that, but use of fixtures in models and libs tests is highly deprecated because fixtures data are not always stable. Use generated objects or factories instead.

burisu:
Replaced by #2292

PR #2033, opened 2017-07-24
Fixed interpolation error in eng locales for payslips#edit (#2032)

Zank94:


Aquaj:
Made obsolete by #2039 (which includes it).

PR #2035, opened 2017-07-24
Removed unknown key :name from intervention locales (#2034)

Zank94:


Aquaj:
Made obsolete by #2039 (which includes it).

PR #2043, opened 2017-07-25
set nature to other_regularization if not in salesaffair and purchase…

### LINKED ISSUE #1888: debt_transfers#create (RuntimeError) "can't run a debt transfer with homogeneous affairs"
## Error
A RuntimeError occurred in debt_transfers#create:

  can't run a debt transfer with homogeneous affairs
  app/models/debt_transfer.rb:71:in `block in <class:DebtTransfer>'

---
## Environment
- **Version :** 2.43.2
- **Server :** production
- **Date :** 13/05/2017

pratap477:
…saffair
Fixes #1888.

### DIFF
--- a/app/models/debt_transfer.rb
+++ b/app/models/debt_transfer.rb
@@ -71,7 +71,8 @@ class DebtTransfer < Ekylibre::Record::Base
                     elsif debt_transfer_affair.is_a?(SaleAffair) && affair.is_a?(PurchaseAffair)
                       :purchase_regularization
                     else
-                      raise 'Cannot run a debt transfer with homogeneous affairs'
+                      # raise 'Cannot run a debt transfer with homogeneous affairs'
+                      :other_regularization
                     end
   end
 


app/models/debt_transfer.rb
```diff
@@ -71,7 +71,8 @@ class DebtTransfer < Ekylibre::Record::Base
                     elsif debt_transfer_affair.is_a?(SaleAffair) && affair.is_a?(PurchaseAffair)
                       :purchase_regularization
                     else
-                      raise 'Cannot run a debt transfer with homogeneous affairs'
```
Aquaj: This simply can't work. The nature of `DebtTransfer`s is enumerized to only accept `:sale_regularization` or `:purchase_regularization` as value. This would directly raise a validation error when run.
The fix must be implemented at a higher scope, the users should simply not be able to try and create a debt transfer between two affairs of the same kind.

Aquaj:
Doesn't fix the issue and introduces erroneous behaviour.

PR #2045, opened 2017-07-25
Handle if record not found during sale credit creation

### LINKED ISSUE #1891: sale_credits#create (NoMethodError) "undefined method `invoice?' for nil:NilClass"
## Error
A NoMethodError occurred in sale_credits#create:

  undefined method `invoice?' for nil:NilClass
  app/models/sale.rb:200:in `block in <class:Sale>'

---
## Environment
- **Version :** 2.43.2
- **Server :** production
- **Date :** 11/05/2017

pratap477:
Fixes #1891.

### DIFF
--- a/app/models/sale.rb
+++ b/app/models/sale.rb
@@ -197,7 +197,7 @@ class Sale < Ekylibre::Record::Base
   end
 
   before_update do
-    if old_record.invoice?
+    if !old_record.nil? && old_record.invoice?
       self.class.columns_definition.keys.each do |attr|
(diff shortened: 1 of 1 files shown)


app/models/sale.rb
```diff
@@ -197,7 +197,7 @@ class Sale < Ekylibre::Record::Base
   end
 
   before_update do
-    if old_record.invoice?
+    if !old_record.nil? && old_record.invoice?
```
Aquaj: Can you use the `.present?` method instead of the negated-nil check?

PR #2047, opened 2017-07-25
Fixes product nature variants show view when fails due to phyto data error

### LINKED ISSUE #1902: product_nature_variants#show (ActionView::Template::Error) "undefined method `usages' for #<struct Struct::U...
## Error
An ActionView::Template::Error occurred in product_nature_variants#show:

  undefined method `usages' for #<struct Struct::UnknownAgent risks=[nil]>
  app/views/backend/product_nature_variants/show.html.haml:68:in `block (2 levels) in _app_views_backend_product_nature_variants_show_html_haml___1877471167772246861_70037869287140'

---
# Environment
- **Version :** 2.43.2
- **Server :** production
- **Date :** 17/05/2017

pratap477:
Fixes #1902.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -18,6 +18,7 @@
 # Ignore the default SQLite database.
 /db/*.sqlite3
 /db/*.sqlite3-journal
+/db/structure.sql
 
 # Ignore the default first run files
 db/first_runs
--- a/app/views/backend/product_nature_variants/show.html.haml
+++ b/app/views/backend/product_nature_variants/show.html.haml
@@ -47,7 +47,7 @@
 
 
 = cobbles do |c|
-  - if resource.france_maaid? && (agent = Pesticide::Agent.find(resource.france_maaid))
+  - if resource.france_maaid? && (agent = Pesticide::Agent.find(resource.france_maaid.to_s))
     - c.cobble(:phyto_data) do
       - if agent.risks.any?
         - avatars = []


.gitignore
```diff
@@ -18,6 +18,7 @@
 # Ignore the default SQLite database.
 /db/*.sqlite3
 /db/*.sqlite3-journal
+/db/structure.sql
```
Aquaj: This is not acceptable. The db/structure.sql is a fundamental part of the application and cannot be `gitignore`d without creating problems in the near future.

app/views/backend/product_nature_variants/show.html.haml
```diff
@@ -47,7 +47,7 @@
 
 
 = cobbles do |c|
-  - if resource.france_maaid? && (agent = Pesticide::Agent.find(resource.france_maaid))
+  - if resource.france_maaid? && (agent = Pesticide::Agent.find(resource.france_maaid.to_s))
```
Aquaj: This is not a solution, the  `agent` returned by `Pesticide::Agent.find` will still be `Pesticide::Agent::Unknown` and will still raise the "Unknown method: `usages` for Unknown Struct" error.

The solution would rather be to add a :usages attribute on the structure and fill it with an empty array.

Aquaj:
Does not fix the issue.

PR #2054, opened 2017-07-27
Add an option before the form to choose if you show the map or not

thib44:
When you select the type of intervention you have a button where you can choose to show or hide the map in the form.


burisu:
[ekylibre:performance/show-map-option](https://trello.com/c/XOoj1ksr/284-ekylibreperformance-show-map-option)

app/models/intervention.rb
```diff
@@ -394,10 +394,6 @@ def human_activities_names
     activities.map(&:name).to_sentence
   end
 
-  def product_parameters
-    InterventionProductParameter.where(intervention: self)
-  end
-
```
burisu: Not the subject

app/controllers/backend/interventions_controller.rb
```diff
@@ -255,9 +255,17 @@ def new
       from_request = Intervention.find_by(id: params[:request_intervention_id])
       @intervention = from_request.initialize_record if from_request
 
+      @show_map = current_user.preference('interface.interventions.map_show', true)
+
       render(locals: { cancel_url: { action: :index }, with_continue: true })
     end
 
+    def update_map_preference
```
burisu: This action can be more RESTful. Can you consider to create a dedicated controller for `Preference` with a strong validation of preferences which can be updated directly through PATCH requests.

.gitignore
```diff
@@ -93,4 +93,6 @@ Procfile.*
 .gitconfig
 
 custom_plan.rb
-zeus.json
\ No newline at end of file
+zeus.json
+
+private/test/*
```
burisu: Not the subject (and updated on master)

config/locales/eng/action.yml
```diff
@@ -1664,6 +1664,7 @@ eng:
     potassium_concentration: "Potassium concentration" #?
     potential_hydrogen: "Potential hydrogen" #?
     precipitation: "Precipitation"
+    preference: "Preference"
```
burisu: Why this translation?

config/locales/fra/action.yml
```diff
@@ -1810,6 +1810,7 @@ fra:
     potassium_concentration_per_hectare: "Teneur en potassium par hectare" #?
     potential_hydrogen: "Potentiel hydrogène (pH)"
     precipitation: "Précipitations"
+    preference: "Préférence"
```
burisu: Why this translation?

PR #2074, opened 2017-08-02
No more `tenant:restore` in prod env.

Aquaj:
Added a twofold mechanism to prevent the use of `rake tenant:restore` when app is in production.
  1- Added a `raise` call unless an ENV variable is set.
  2- Added 3 more confirmations (for a total of 5 prompts) before allowing the user to overwrite an existing tenant.

In the process of adding the multiple prompts I set up an array of confirmation messages for easy extension/modification instead of successive `confirm` calls.


Aquaj:
[ekylibre:feature/deactivate-restore-in-production-env](https://trello.com/c/5Er805aB/294-ekylibrefeature-deactivate-restore-in-production-env)

PR #2076, opened 2017-08-03
Added a way to make cultivable zones valid if there are crossing lines when drawing shapes (2075)

Zank94:


Zank94:
[ekylibre:debug/gh2075-Fix-cultivable-zones-import-from-first-run](https://trello.com/c/1skzKP2I/295-ekylibredebug-gh2075-fix-cultivable-zones-import-from-first-run)

PR #2079, opened 2017-08-03
Removed start intervention form on each crumb to prevent timeout on intervention_participations index

Zank94:
Temporary solution: commented form displayed on each crumb
Next step: use a modal to prevent timeout while keeping the form
Answer to #2078


Zank94:
[ekylibre:debug/gh2078-Timeout-intervention-participations-index](https://trello.com/c/MukprU1F/296-ekylibredebug-gh2078-timeout-intervention-participations-index)

PR #2084, opened 2017-08-04
FixedAssets that are on closed financial periods

sandy1987:
Hi,

I have put conditions over the fixed-assets section and also update yml file to  display warning message in both EN and FRA


app/views/backend/fixed_assets/show.html.haml
```diff
@@ -1,3 +1,6 @@
+- if resource.started_on <= resource.journal.closed_on
+  .flash.warning
```
Aquaj: Could you rather use the `notify_warning_now` method in the controller #show action rather than inserting it in the view directly?

app/views/backend/base/_state_bar.html.haml
```diff
@@ -10,7 +10,11 @@
       - classes << "active"
       %span{class: classes}= localized_name
     - elsif transition = transitions.detect{|t| t.to == state.name.to_s }
-      = link_to(localized_name, {action: transition.event, id: resource.id}, class: classes, method: :post, data: {confirm: :are_you_sure.tl}, title: transition.event.to_sym.ta)
```
Aquaj: Hello Sandeep, this would technically work I guess but this is unsustainable, we cannot add special cases for each transition we want to disable. Could you take a look at the `state_machine` in the FixedAsset model instead ? You can find an example on how to use a condition on one of the transitions in the `state_machine` block of the model Parcel (line 122), rather than having to modify the html itself.

sandy1987: @Aquaj ,
I have updated the code and made changes according to you. Please take a look and let me know if any thing missing.
Thanks

app/models/fixed_asset.rb
```diff
@@ -193,6 +193,11 @@ class FixedAsset < Ekylibre::Record::Base
     depreciate! if @auto_depreciate
   end
 
+  def any_financial?
+    self.started_on <= self.journal.closed_on ? false : true
+  end
+
+
```
Aquaj: Could you remove the useless blank line?

app/controllers/backend/fixed_assets_controller.rb
```diff
@@ -100,6 +100,7 @@ def show
 
       return unless @fixed_asset = find_and_check
       t3e @fixed_asset
+      notify_warning_now(:closed_financial_periods) if (@fixed_asset.started_on <= @fixed_asset.journal.closed_on) rescue false
```
Aquaj: Could the check in the condition be replaced with a call to the method that is also checking that in order to have [DRY](https://en.wikibooks.org/wiki/Ruby_on_Rails/Getting_Started/Don%C2%B4t_repeat_yourself)-er code ?

app/models/fixed_asset.rb
```diff
@@ -193,6 +193,11 @@ class FixedAsset < Ekylibre::Record::Base
     depreciate! if @auto_depreciate
   end
 
+  def any_financial?
+    self.started_on <= self.journal.closed_on ? false : true
```
Aquaj: The return value must match the name. If the name is `any_financial?` or `on_closed_periods?` it should return true when it is on closed periods or when there are financial_years closed on the duration. Right now the return value is the opposite of what would be expected by reading the name.
Either the name of the method needs to change to match the behavior, or the behavior must change to match the method name. (Note: it is possible to use `transition ... unless: <method>'  just like you used the `transition ... if: <method>` in the state_machine declaration.)

app/models/fixed_asset.rb
```diff
@@ -193,6 +193,11 @@ class FixedAsset < Ekylibre::Record::Base
     depreciate! if @auto_depreciate
   end
 
+  def any_financial?
```
Aquaj: Could we find a more explicit method name? Something like 'on_closed_periods?' maybe.

app/models/fixed_asset.rb
```diff
@@ -193,6 +193,10 @@ class FixedAsset < Ekylibre::Record::Base
     depreciate! if @auto_depreciate
   end
 
+  def on_unclosed_periods?
+    self.started_on <= self.journal.closed_on ? false : true
```
Aquaj: The ternary operator feels unnecessary here. Changing the operator (`>` is logically symmetrical to `<=`) would allow us to remove it, making for a cleaner method body.

PR #2093, opened 2017-08-08
Improve performance of plants#index, add jsonb column to save surface_area

thib44:


app/models/concerns/indicateable.rb
```diff
@@ -65,8 +65,7 @@ def reading(indicator, options = {})
       raise ArgumentError, "Unknown indicator #{indicator.inspect}. Expecting one of them: #{Nomen::Indicator.all.sort.to_sentence}."
     end
     read_at = options[:at] || Time.zone.now
-    indicator_name = indicator.name
-    readings.where(indicator_name: indicator.name).where('read_at <= ?', read_at).order(read_at: :desc).first
+    readings.where("indicator_name = ? AND read_at <= ?", indicator.name, read_at).max_by(&:read_at)
```
burisu: `max_by` adds performance issue, because it requires all data of query before sorting them. `.order(...).first` limits data to one row only directly.

app/models/product.rb
```diff
@@ -763,21 +775,36 @@ def variables(_options = {})
...
+  def net_surface_area
+    computed_surface = self.reading_cache[:net_surface_area] || self.reading_cache['net_surface_area']
+    return computed_surface if computed_surface
+    self.net_surface_area = calculation_of_net_surface_area
+  end
+
   # Override net_surface_area indicator to compute it from shape if
   # product has shape indicator unless options :strict is given
-  def net_surface_area(options = {})
+  def calculation_of_net_surface_area(options = {})
```
burisu: Please consider to rename method to `calculate_net_surface_area`. An action verb at start is better.

app/models/product_reading.rb
```diff
@@ -90,6 +90,20 @@ class ProductReading < Ekylibre::Record::Base
     end
   end
 
+  after_save do
+  last_reading = product.readings.where(indicator_name: indicator_name).last
+    if last_reading.read_at <= read_at
+      product.send("#{indicator_name}=", value)
+    end
+
+
```
burisu: Too many blank lines

app/views/backend/plants/index.html.haml
```diff
@@ -23,4 +23,5 @@
       = plants_map main: true
 
   - j.face :list do
-    = main_list
+    - Ekylibre.profile('plants_list') do
```
burisu: No `profile` in PR

test/models/product_test.rb
```diff
@@ -94,4 +94,15 @@ class ProductTest < ActiveSupport::TestCase
...
     end
   end
+
+  test 'sets' do
+    p = Plant.last
+    # p.id = nil
+    # p.number = nil
+    # prod = Plant.new(p.attributes)
+    # assert_equal true, prod.valid?
+    binding.pry
```
burisu: No `binding.pry` in PR

app/models/product_reading.rb
```diff
@@ -90,6 +90,19 @@ class ProductReading < Ekylibre::Record::Base
     end
   end
 
+  after_save do
+    last_reading = product.readings.where(indicator_name: indicator_name).last
+    if last_reading.read_at <= read_at
+      product.send("#{indicator_name}=", value)
+      product.update_column(:reading_cache, product.reading_cache)
```
burisu: This way to do can generate errors on concurrent calls

Aquaj: Have you an idea of how we could avoid these errors? Should I add a semaphore call around it to ensure it won't get weird with concurrency ?

PR #2095, opened 2017-08-08
Split factories file into separate files for each model

Zank94:


Zank94:
[ekylibre:factory-girl/split-factories-file-into-separate-files](https://trello.com/c/yv4wnw1Q/301-ekylibrefactory-girl-split-factories-file-into-separate-files)

PR #2096, opened 2017-08-08
Fixes performance activity show view issue

Aquaj:
Did some heavy rewrite of the `caliber_series` method and some refacto of the view and other helpers.

The performance is slightly better but still not great although I have a hard time figuring out the culprit since the logs show vastly different values in prod vs local. (7000ms in view + 700ms in ActiveRecord for the local rendering, 4000ms in view, 4500ms in AR on the production server).


burisu:
Rubocop can help you to fix codacy issues

PR #2105, opened 2017-08-16
Add new method to permit ratio for activities reporting

ionosphere:


app/views/backend/land_parcels/show.html.haml
```diff
@@ -6,9 +6,10 @@
   - c.cobble :general_informations do
     = product_mini_map
     = infos do
-      .med-info
-        %span.title= LandParcel.human_attribute_name(:population)
-        %span.value= resource.population.round(3).l + " " + resource.unit_name
+      - if resource.initial_shape
+        .med-info
+          %span.title= LandParcel.human_attribute_name(:net_surface_area)
+          %span.value= resource.initial_shape_area.round(2).in(:hectare)
```
burisu: I understand the why but, we can't put `:hectare` everywhere like that, this won't work for small land parcels which are in square meters. I propose to use the `Activity#size_unit` to define which unit to use. A delegation should be added in `LandParcel` model or a method like `human_initial_shape_area` that will return good value with good unit.

app/models/intervention.rb
```diff
@@ -575,6 +575,22 @@ def working_area(unit = :hectare)
     ActiveSupport::Deprecation.warn 'Intervention#working_area is deprecated. Please use Intervention#working_zone_area instead.'
     working_zone_area(unit)
   end
+  
+  def ratio_for(activity)
```
burisu: Method name is not very explicit. Change it or add a comment. I prefer largely the first choice.

app/models/intervention.rb
```diff
@@ -575,6 +575,22 @@ def working_area(unit = :hectare)
...
+      precision = 2
+      if targets.any?
+        at = targets.of_activity(activity).with_working_zone.map(&:working_zone_area).sum.in(unit)
+        coeff = (at.to_d / working_zone_area.to_d) if working_zone_area.to_d != 0.0
+        if coeff
+          return coeff.round(precision)
+        else
+          return 1.00
+        end
+      end
```
burisu: What about the `else`s for given lines 580 and 583?

app/models/intervention.rb
ionosphere: OK but what's the first choice ?

app/models/intervention.rb
ionosphere: begin with instantiate coeff = 1.00 is more suitable instead of the else ?

app/models/intervention.rb
burisu: 1. Rename the method
2. Add an explicit comment

app/models/intervention.rb
burisu: Surely, I don't look at it precisely. Better than nil I hope.

app/models/intervention.rb
burisu: Or activity_cost_ratio_for?

PR #2110, opened 2017-08-17
Debug/2017 crops are not updated when planting

sandy1987:
I've resolved this update issue( Issue-id). During debug I've find out that their is an issue while changing the "Land parcel option" on the edit page of intervention.
If we update any working zone shape of land parcel, it gets successfully saved. But on change of land parcel from the option list, it again shows the default shape of working zone while shows correct updated shape on reloading the page.


Aquaj:
Obsolete.

PR #2114, opened 2017-08-18
Fixed purchase and sale label field to match potential item variant update

### LINKED ISSUE #2111: Error in invoice duplication
When an invoice is duplicated and the variant is changed (Ex : Vin -> Aliment en vrac) 
capture du 2017-08-17 14-53-02

The variant is not modified on the invoice (Vin): 
capture du 2017-08-17 14-53-21

But if we make an asset of this last invoice, the variant is well modified (Aliment en vrac) : 
capture du 2017-08-17 14-53-59

Zank94:
Fixes #2111


Zank94:
[ekylibre:debug/gh#2111-Fix-variant-name-displayed-after-updating-sale-item](https://trello.com/c/p4qlZEFN/314-ekylibredebug-gh2111-fix-variant-name-displayed-after-updating-sale-item)

PR #2142, opened 2017-08-25
Add autocomplete to journal entry form

Zank94:


Zank94:
[ekylibre:feature/accountancy-2018/add-autocomplete-to-journal-entry-form](https://trello.com/c/9moWfxLV/332-ekylibrefeature-accountancy-2018-add-autocomplete-to-journal-entry-form)

app/assets/javascripts/ui.js
```diff
@@ -198,38 +198,55 @@
...
+  // Nullify sibling input
   $(document).on("keyup", "input[data-exclusive-nullify]", function () {
     var element = $(this);
-    var scope = $("html");
-    if (element.data("use-closest")) {
-      scope = element.closest(element.data("use-closest"));
-    }
     if (element.numericalValue() !== 0) {
-      scope.find(element.data("exclusive-nullify")).val('');
+      $(element).closest('tr').find('input'+element.data('exclusive-nullify')).val('');
```
Aquaj: This should be 're-generified', (see old implementation that uses `data-use-closest` rather than `closest(tr)`). This also begs the question of why it was rewritten.

app/assets/javascripts/ui.js
```diff
@@ -198,38 +198,55 @@
...
     }
   });
 
+  // Nullify inputs if it filled
```
Aquaj: We do not need to keep commented code! This is what Git is for :wink:

PR #2162, opened 2017-08-30
Debug/gh2157 product movements not deleted when a parcel is destroyed

### LINKED ISSUE #2157: Product movements not deleted when a parcel is destroyed
Population value and movements related to a specific variant remain unchanged after a parcel (state: given) containing this variant as parcel item is destroyed.
screenshot at 2017-08-30 11 15 24

Zank94:
Fixes #2157


Zank94:
[ekylibre:debug/gh2157-product-movements-not-deleted-when-a-parcel-is-destroyed](https://trello.com/c/r46yKYgm/335-ekylibredebug-gh2157-product-movements-not-deleted-when-a-parcel-is-destroyed)

app/models/parcel_item.rb
```diff
@@ -70,6 +66,11 @@ class ParcelItem < Ekylibre::Record::Base
   has_one :delivery, through: :parcel
   has_one :storage, through: :parcel
   has_one :contract, through: :parcel
+  has_one :product_enjoyment, class_name: "ProductEnjoyment", as: :originator, dependent: :destroy
```
Aquaj: You can remove the "class_name" argument as the relations are already named like the classes they link to.

app/models/parcel_item.rb
```diff
@@ -206,8 +207,16 @@ def name
   # Mark items as given, and so change enjoyer and ownership if needed at
   # this moment.
   def give
-    give_outgoing if parcel_outgoing?
-    give_incoming if parcel_incoming?
+    if parcel_outgoing?
```
Aquaj: Could you replace this by a single transaction block ?
```
transaction do
  give_outgoing if..
  give_incoming if..
end
```

test/models/parcel_item_test.rb
```diff
@@ -54,5 +54,20 @@
 
 class ParcelItemTest < ActiveSupport::TestCase
   test_model_actions
-  # Add tests here...
+
+  test "give_incoming doesn't create ProductMovement and ProductLocalization if an error occurs when creating ProductEnjoyment"  do
```
Aquaj: Could you rename this test ? You're testing a broader behaviour: that give doesn't create the dependent records if there's an exception during the process. Not specifically ProductMovement/Localization (even if that's what you used to test said behaviour).

test/models/parcel_item_test.rb
```diff
@@ -54,5 +54,20 @@
...
+  test "give doesn't create the dependent records if there is an exception during the process" do
+    product = create(:product)
+    parcel_item = create(:parcel_item, product: product, product_identification_number: '12345678', product_name: 'Product name')
+    ProductMovement.destroy_all
+    ProductLocalization.destroy_all
+    ProductEnjoyment.stub :create!, lambda { |*args| raise } do
+      begin
+        parcel_item.give
+      rescue
+        puts "Error message"
```
Aquaj: Can you remove this `puts` that will not make much sens when looking through the test logs ?

PR #2169, opened 2017-09-01
#feature changes on the crops name after seeding

sandy1987:
I have update "New name" if its not present. According to given format

### DIFF
--- a/app/models/intervention.rb
+++ b/app/models/intervention.rb
@@ -478,9 +478,10 @@ class Intervention < Ekylibre::Record::Base
         stopped_at: self.stopped_at
       )
     end
-    outputs.find_each do |output|
+    outputs.each_with_index do |output,index|
       product = output.product
       next unless product
+      set_product_name(product,index) if product.name == "blank"
       product.born_at = self.started_at
       product.initial_born_at = product.born_at
       product.save!
@@ -508,6 +509,13 @@ class Intervention < Ekylibre::Record::Base
       movement.stopped_at = self.stopped_at
       movement.save!
     end
+  end
+
+  # Set product name if its blank
+  def set_product_name(product,index)
+    land_parcel_name = self.product_parameters.where(type: "InterventionTarget", reference_name: "land_parcel")[index].product.name rescue ""
+    culture_name = self.product_parameters.where(:type=> "InterventionOutput",reference_name: "plant")[index].variant.name rescue ""
+    product.name  = "#{land_parcel_name} #{culture_name} #{self.created_at.strftime('%Y')} #{self.id}" if (!land_parcel_name.blank? && !culture_name.blank?)
   end
 
   # Sums all intervention product parameter total_cost of a particular role
--- a/app/models/intervention.rb
+++ b/app/models/intervention.rb
@@ -478,7 +478,7 @@ class Intervention < Ekylibre::Record::Base
         stopped_at: self.stopped_at
       )
     end
-    outputs.find_each  do |output|
+    outputs.find_each do |output|
       product = output.product
       next unless product
       product.born_at = self.started_at
--- a/app/models/intervention.rb
+++ b/app/models/intervention.rb
@@ -481,7 +481,6 @@ class Intervention < Ekylibre::Record::Base
     outputs.each_with_index do |output,index|
       product = output.product
       next unless product
-      set_product_name(product,index) if product.name == "blank"
       product.born_at = self.started_at
       product.initial_born_at = product.born_at
       product.save!
@@ -509,13 +508,6 @@ class Intervention < Ekylibre::Record::Base
       movement.stopped_at = self.stopped_at
       movement.save!
     end
-  end
-
-  # Set product name if its blank
-  def set_product_name(product,index)
-    land_parcel_name = self.product_parameters.where(type: "InterventionTarget", reference_name: "land_parcel")[index].product.name rescue ""
-    culture_name = self.product_parameters.where(:type=> "InterventionOutput",reference_name: "plant")[index].variant.name rescue ""
-    product.name  = "#{land_parcel_name} #{culture_name} #{self.created_at.strftime('%Y')} #{self.id}" if (!land_parcel_name.blank? && !culture_name.blank?)
   end
 
   # Sums all intervention product parameter total_cost of a particular role
--- a/app/models/intervention.rb
+++ b/app/models/intervention.rb
@@ -478,7 +478,7 @@ class Intervention < Ekylibre::Record::Base
         stopped_at: self.stopped_at
       )
     end
-    outputs.each_with_index do |output,index|
+    outputs.find_each  do |output|
       product = output.product
       next unless product
       product.born_at = self.started_at
--- a/app/models/intervention_output.rb
+++ b/app/models/intervention_output.rb
@@ -76,7 +76,7 @@ class InterventionOutput < InterventionProductParameter
       reading = readings.find_by(indicator_name: :shape)
       output.initial_shape = reading.value if reading
       output.save!
-
+      choose_default_name(output)
       if intervention.record?
         movement = product_movement
         movement = build_product_movement(product: output) unless movement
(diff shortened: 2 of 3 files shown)


app/models/intervention.rb
```diff
@@ -510,6 +511,13 @@ def update_temporality
     end
   end
 
+  # Set product name if its blank
+  def set_product_name(product,index)
```
Aquaj: Instead of using 'index' which adds a gigantic amount of coupling, you should rely on the fact that "land_parcel" and "plant" share the same `InterventionGroup` (ex: "output.group.targets.first #=> land_parcel").

app/models/intervention.rb
```diff
@@ -510,6 +511,13 @@ def update_temporality
...
+    land_parcel_name = self.product_parameters.where(type: "InterventionTarget", reference_name: "land_parcel")[index].product.name rescue ""
+    culture_name = self.product_parameters.where(:type=> "InterventionOutput",reference_name: "plant")[index].variant.name rescue ""
```
Aquaj: In such a case pass in the output as a parameter instead of relying on indexes.

app/models/product.rb
```diff
@@ -498,6 +498,7 @@ def set_initial_values
   # Try to find the best name for the new products
   def choose_default_name
     return if name.present?
+    return self.name = "blank" if (name.blank? && self.class == Product)
```
Aquaj: Why would you set the name to the literal string "blank"? Also what is is `== Product` in the condition for?

app/models/intervention.rb
sandy1987: @Aquaj ,
Actually I used .first concept, But Here add more options their user can select culture and land parcel multiple times, So .first is not working, That the reason I used Index here,

app/models/intervention.rb
sandy1987: @Aquaj ,

Here Multiple land parcel and culture present, Whose new name field are blanked then we have to update the name based on land parcel and culture name.

app/models/product.rb
sandy1987: @Aquaj ,
Reason behind this when we create seeding or planting this method call multiple times, That the reason I have used "== Product" to check the Object, And I have to set "blank"  because we get this blank value on the other model, So we easy to get this value and update accordingly.

sandy1987:
Hello @Aquaj 
Under the "Parameters" section, User click on the "Add Zone", then he will able to select multiple land parcel and culture.
For Example:-
If user set multiple land parcel and culture, and both the Zone he will not fill "New Name" field.
1) Land Parcel and Culture 
land parcel = La Michardière
culture  = Culture de vigne
intervention_id = 116
2) Land Parcel and culture
land parcel = La Boutanche
culture  = Culture de blé
intervention_id = 116
Please suggest me which value set default Under the "New Name".
1) La Michardière Culture de vigne 2017 116
2) La Boutanche Culture de blé 2017 116

Please suggest which value I will set 1 or 2.
And If we used object.first  then the "New Name" comes same on both the places for above example
our "New Name" value is "La Michardière Culture de vigne 2017 116" or "La Boutanche Culture de blé 2017 116"

app/models/intervention.rb
Aquaj: In the case of a sowing intervention like here the structure is : 
```
Intervention --(many)-> InterventionGroup   --> InterventionTarget
                                             \-->  InterventionOutput
```
The association you use goes straight ignoring the groups to the outputs and targets, like :
```
Intervention --(many)-> InterventionOutputs
              \--(many)-> InterventionTargets
```
So when you are manipulating an `InterventionOutput` you can use that grouping hierarchy:
```
[ output ]
             Intervention
           /             \
       IGroup           IGroup
     /       \         /      \
> IOutput  ITarget    IOutput  ITarget
```
```
[ output.group ]
             Intervention
           /             \
    >  IGroup           IGroup
     /       \         /      \
 IOutput  ITarget    IOutput  ITarget
```
```
[ output.group.targets ]
             Intervention
           /             \
       IGroup           IGroup
     /       \         /      \
 IOutput  [ITarget]   IOutput  ITarget
```
```
[ output.group.targets.first ]
             Intervention
           /             \
      IGroup           IGroup
     /       \         /      \
 IOutput  ITarget <   IOutput  ITarget
```

What you refer to must be the `outputs` and `targets` on the `Intervention`, not on the `InterventionGroup` where you were using a structure looking like :
```
[ output ]
        Intervention
      ___ |      |_____
     /    |      |     \
    /     |      |      \
   /      |      |       \
>IOutput IOutput ITarget ITarget
```
```
[ output.intervention ]
      > Intervention
      ___ |      |_____
     /    |      |     \
    /     |      |      \
   /      |      |       \
IOutput IOutput ITarget ITarget
```
```
[ output.intervention.targets ]
        Intervention
      ___ |      |_____
     /    |      |     \
    /     |      |      \
   /      |      |       \
IOutput IOutput [ITarget ITarget]
```
```
[ output.intervention.targets.first ]
      > Intervention
      ___ |      |_____
     /    |      |     \
    /     |      |      \
   /      |      |       \
IOutput IOutput ITarget< ITarget 
```

app/models/intervention.rb
Aquaj: See my answer to the previous comment. I detail the structure and how you can avoid to have to do so.

Aquaj:
The name should be a) for the first Culture and b) for the second one just like you thought. I have detailed the existing architecture that allows this in [one of my comments](https://github.com/ekylibre/ekylibre/pull/2169#discussion_r136943881).

app/models/product.rb
Aquaj: I am afraid I still don't really get what you mean.

app/models/product.rb
```diff
@@ -498,6 +498,7 @@ def set_initial_values
   # Try to find the best name for the new products
   def choose_default_name
     return if name.present?
+    return self.name = "blank" if name.blank?
```
Aquaj: Why don't you just leave the name blank?

Aquaj: Why don't you just leave the name blank?

app/models/intervention_output.rb
```diff
@@ -93,6 +93,15 @@ class InterventionOutput < InterventionProductParameter
     end
   end
 
+  def choose_default_name(output)
+    if output.name.eql?("blank")
+      land_parcel_name = self.group.targets.first.product.name rescue ""
```
Aquaj: Could you replace the `rescue`s by conditionals? It is more explicit and will facilitate debug later on.

app/models/intervention_output.rb
```diff
@@ -93,6 +93,15 @@ class InterventionOutput < InterventionProductParameter
...
+      culture_name = self.variant.name rescue ""
```
Aquaj: Could you replace the `rescue`s by conditionals? It is more explicit and will facilitate debug later on.

app/models/intervention_output.rb
```diff
@@ -93,6 +93,15 @@ class InterventionOutput < InterventionProductParameter
...
+      output.name = "#{land_parcel_name} #{culture_name} #{self.group.intervention.created_at.strftime('%Y')} #{self.group.intervention.id}"
```
Aquaj: This should be the intervention.number and not id as the user won't recognize those.

app/models/product.rb
sandy1987: Hello,
I have try to leave blank name, But name validation presence: true, So  we are not able to create Product , And Here I am not find interventions and its details so unable to set the name value that the reason I will pass "blank" text and On the InterventionOutput I get this name and update its value accordingly,

app/models/intervention_output.rb
sandy1987: In place of this we used self.group.targets.try(:first).try(:product).try(:name),
Because in this case try handle exception if its occur

sandy1987:
@Aquaj 
I have just update the code please review and let me know if any changes required.
thanks

app/models/product.rb
```diff
@@ -498,6 +498,7 @@ def set_initial_values
   # Try to find the best name for the new products
   def choose_default_name
     return if name.present?
+    return if name.blank?
```
Aquaj:
```
return if name.present?
return if name.blank?
```
is equivalent to have deleted replaced the whole method by a simple `return`, this would break behavior in all the other places this method is called.

sandy1987: @Aquaj 
Here Previously I used self.class == Product  This condition return  true in our case only, when we create the Product, Can I used this?
So our new condition will be like that

return if name.present?
return if name.blank? && self.class == Product

Its okay

PR #2200, opened 2017-09-19
show the map by default and correction of translation

thib44:


app/helpers/backend/base_helper.rb
```diff
@@ -478,7 +478,7 @@ def chronology_period_icon(positioned_at, picto_class, html_options = {})
 
     def user_preference_value(name)
       preference = current_user.preferences.find_by(name: name)
-      preference ? preference.value : nil
+      preference ? preference.value : true
```
Aquaj: Could you change it only via the `default_value` parameter on the `#get` call of your preference instead of changing it for *all preferences* ?

thib44: Ok I've juste changed it, by adding an default argument to `user_preference_value` and adding a default value in the view

PR #2203, opened 2017-09-19
Performance/cartography speeding up

Nyxelio:
Loads land parcel and plant layers asynchronously in map on intervention form.


Nyxelio:
Yes we already have a spinner to indicate shapes are loading, on the bottom left.

PR #2213, opened 2017-09-27
Async loads stock in ground on activity show view

Nyxelio:
Implements async cobble to load stock in ground asynchronously and speed up page rendering.
Depends #2205


Aquaj:
Waiting on #2205 to review.

PR #2216, opened 2017-09-27
add a callback to block redirection to backend if the first-run is no…

thib44:
…t finish

### DIFF
--- a/app/controllers/backend/base_controller.rb
+++ b/app/controllers/backend/base_controller.rb
@@ -32,6 +32,7 @@ module Backend
     before_action :set_current_period_interval
     before_action :set_current_period
     before_action :publish_backend_action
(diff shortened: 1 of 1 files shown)


thib44:
https://github.com/ekylibre/ekylibre-hajimari/pull/10

app/controllers/backend/base_controller.rb
```diff
@@ -32,6 +32,7 @@ class BaseController < ::BaseController
     before_action :set_current_period_interval
     before_action :set_current_period
     before_action :publish_backend_action
+    before_action :first_run_or_backend
```
Aquaj: This can't be in the core of Ekylibre, as all Ekylibre install won't have the Hajimari plugin.

PR #2223, opened 2017-10-09
Add an alert when the date is not in the current period

thib44:


app/helpers/backend/base_helper.rb
```diff
@@ -292,6 +292,11 @@ def financial_year_selector(financial_year_id = nil, options = {})
       render 'backend/shared/financial_year_selector', financial_year: current_user.current_financial_year, param_name: options[:param_name] || :current_financial_year
     end
 
+    def financial_year_started_on_stopped_on
```
burisu: Instead of using an helper that only provide a hash to include in HAML tag which is used by JS. 
Use pure `data-` attributes on the targeted input: `data-alert-if-out-of-period` (with alert message as value), `data-period-started-on` (with start date as value) and `data-period-stopped-on` (with finish date as value).
You can have an helper (`current_financial_year_inclusion_attributes`) which will return the attributes directly for `input_html` option of simple form input.
Your JS will detect attributes and build wanted HTML on alert or before (like you prefer).

app/themes/tekyla/stylesheets/forms.scss
```diff
@@ -383,4 +383,9 @@ div.nested-fields {
   padding-left: $default-gap;
   padding-right: $default-gap;
   @include inline-block;
-}
\ No newline at end of file
+}
+
+#financial-year-dates {
+  color: #e61e1a;
+  margin-left: 45px;
```
burisu: Use SCSS variable in order to be adaptable

config/locales/fra/action.yml
```diff
@@ -2304,6 +2304,7 @@ fra:
       resized: "Mode «\_Tout à l’écran\_»" #?
     virtual_quantity: "Quantité virtuelle" #?
     wages: "Rémunérations"
+    warning_date_not_in_period: "Attention : La date saisie n'est pas sur l'exercice en cours."
```
burisu: Adds english translation first and then `rake clean` to update all translations.

PR #2226, opened 2017-10-11
Translation on hover text of time picker

### LINKED ISSUE #1464: Missing translation: hover text on time picker
On the new time picker hover text isn't translated

chetannibotechnologies:
Fixes #1464.


config/locales/cmn/i18n.rb
```diff
@@ -3,7 +3,7 @@
 { cmn: {
   i18n: {
     dir: 'ltr',
-    iso2: 'zh',
+    iso2: 'zh-cn',
```
burisu: Why do you change Chinese `ISO2` code (internal code for translation for _ISO639 code in 2 characters_)?
`zh` is the official code (https://en.wikipedia.org/wiki/Chinese_language). Do you encountered any trouble with JS i18n lib? Instead of ISO2, we may try to change reference code to IETF one which best suit here.
In all case, `iso2` code is expected to be on 2 characters only.

burisu: If nothing blocking with 2-characters code, only revert this change please.

PR #2227, opened 2017-10-12
Validation Failed error fixed

### LINKED ISSUE #1937: Journal entries of loan repayments are not generated
From the /backend/loans view, clicking on "Passer en comptabilité" should generate journal entries for the loan repayments concerned. But it doesn't work anymore

chetannibotechnologies:
Fixes #1937.


burisu:
Hi @chetannibotechnologies!
I don't understand how you find this solution to this bug I can't reproduce. How have you found that?

chetannibotechnologies:
I found this error while accounting loan on my local.

PR #2230, opened 2017-10-17
update cost_amount_computation in InterventionAgent

thib44:


EmmaDorie:
@burisu, this correction is really important for users.
@thib44 no longer works on emergencies, but a solution should be found for merged it in the next version.
Thanks

PR #2232, opened 2017-10-18
Performance/put production cost in cache

Zank94:


Aquaj:
Stale, needs review & potential rework

PR #2235, opened 2017-10-24
translation missing fix

chetannibotechnologies:


config/routes.rb
```diff
@@ -378,7 +378,7 @@
       end
     end
 
-    resources :cultivable_zones, concerns: %i[list unroll], path: 'cultivable-zones' do
+    resources :cultivable_zones, concerns: %i[list unroll], path: 'cultivable_zones' do
```
burisu: Please restore back the path `cultivable-zone` except there is a specific reason to change it.

chetannibotechnologies: Work number link doesn't work because of defined path cultivable-zones in routes, So I changed it to cultivable_zones and its working correctly.

Issue #2256, opened 2017-11-07
Purchase-process : error when we try to display an entity (contact or organization)


EmmaDorie:
Several issues seems to be involved ...

thib44:
https://github.com/ekylibre/ekylibre/pull/2259

PR #2259, opened 2017-11-08
access to entity show ok

thib44:


thib44:
https://github.com/ekylibre/ekylibre/issues/2256

Issue #2266, opened 2017-11-09
Purchase-process : error when we try to click on the number of purchase in the cell "Last purchases" in "purchases" dashboard


EmmaDorie:
![tdb_achats](https://user-images.githubusercontent.com/11627573/32609190-13aaa098-c55f-11e7-9691-c6dcd1e78cc3.png)

PR #2267, opened 2017-11-09
Feature/purchase process/improve purchase dashboard with orders and invoices

thib44:


thib44:
https://github.com/ekylibre/ekylibre/issues/2266

Issue #2263, opened 2017-11-08
Purchase-process : error when try to display a "product_nature_variants"


EmmaDorie:
![variantes](https://user-images.githubusercontent.com/11627573/32561658-15b12824-c4ad-11e7-8870-312118378ee7.png)

thib44:
https://github.com/ekylibre/ekylibre/pull/2271/files

PR #2271, opened 2017-11-09
add reception in belongs to of purchase item

thib44:


thib44:
https://github.com/ekylibre/ekylibre/issues/2263

PR #2278, opened 2017-12-05
Generate continuous number on entries confirm

lcoq:


ionosphere:
@burisu it's ok for me, could you check and validates this pull request ?

burisu:
@ionosphere, I reviewed code too fast. I think that the choosen manner to do can provide errors in numeration: as sequence "current number" is never restored back in failed transaction, it's possible to have a SQL transaction which failed which change sequence current number without any entries updates. It's what we clearly don't want.

Moreover, I some other changes to propose:

- `continuous_number` is not clear, if the numeration is continuous, not the number. I propose to rename current `number` into `document_number` (or `accounting_document_number`), and create a new `number` column which will store these unique values
- Use a DB flag to lock journal entries validation.
- Merge second migration into the first

ionosphere:
Probably more Clear, it could be done

PR #2281, opened 2017-12-07
Fix bad computation of cost per area when working zone is leather tha…

### LINKED ISSUE #2231: the graph of production costs in activity / production is false
when several crops are on a production, the average costs are added instead of being prorated on the surface of each crop.
Ex : if you seed 2 crops on a production (3ha) :  
- 2ha with 100€/ha of seed cost
- 1ha with 90€/ha of seed cost

The graph displays 190€/ha of seed cost

The chart should show (100x2+90x1)/3=96.66€/ha

ionosphere:
Fixes #2231.


burisu:
Tests are failing. Needs a fix.

PR #2282, opened 2017-12-11
Entries continuous number fixes

lcoq:
* Merge `JournalEntries#validated_at` migration with `JournalEntries#continuous_number`
* Use last `JournalEntries#continuous_number` to compute the new one to prevent gaps with rollback


lcoq:
@burisu I've updated this PR to prevent concurrency issue while generating the continuous number : 

* To ensure sequence will never break, I've added a unique constraint on the `continuous_number` column
* Entries validations now locks the entire `journal_entries` table in exclusive mode to prevent concurrent runs

PR #2292, opened 2018-01-09
Weighted average unit cost for stocks

burisu:


Aquaj:
Stale, needs review & rework

PR #2297, opened 2018-06-20
When creating a new account, number can't start with a "0" (I.E: 0123…

DucMirack:
…45) or have all numbers after the first number to be "0" (I.E: 200000)


app/models/account.rb
```diff
@@ -540,6 +544,13 @@ def journal_entry_items_calculate(column, started_at, stopped_at, operation = :s
     journal_entry_items.where(printed_on: started_at..stopped_at).calculate(operation, column)
   end
 
+  def is_number_valid?
+    is_valid = true
+    split_number = self.number.split("")
+    is_valid = false if split_number.first == "0" || split_number.drop(1).all? { |n| n == "0" }
```
Aquaj: I don't see how this prevents from writing in the radix accounts (1, 2, 3, 4, 5, 6, 7, 8) ?

app/models/account.rb
```diff
@@ -540,6 +544,13 @@ def journal_entry_items_calculate(column, started_at, stopped_at, operation = :s
     journal_entry_items.where(printed_on: started_at..stopped_at).calculate(operation, column)
   end
 
+  def is_number_valid?
```
Aquaj: Why not use a something like:
```
def is_number_valid?
  digits = self.number.split
  <your condition>
end
```
Instead of using unnecessary variables and a `false if` ?

Issue #82, opened 2012-12-11
Add paper_trail by default in all models


burisu:
Via CompanyRecord ?

burisu:
Version model is our solution for now.

PR #2311, opened 2018-08-14
UX Debug #82 date popup validation problem

burisu:


Aquaj:
Fixed elsewhere (flatpicker)

PR #2313, opened 2018-08-28
Removed attr_readonly on account number which prevented user from upd…

Zank94:
…ating it


Aquaj:
Are we sure we want users to be able to update account numbers? Isn't that risky ?

Zank94:
According to @AnneKappy it should be allowed, correct me if I'm wrong

PR #2314, opened 2018-09-06
Implemented optional n-1 display for balance

Zank94:


config/locales/fra/action.yml
```diff
@@ -1248,6 +1248,7 @@ fra:
     disease_percentage_evolution: "Evolution des maladies"
     display: "Afficher"
     display_help: "Afficher l’aide" #?
+    display_n_1: "Afficher N-1"
```
Aquaj: Do we really want "N-1" and not something like "période précédente" ?

config/locales/eng/action.yml
```diff
@@ -1302,6 +1302,7 @@ eng:
     disease_percentage: "Disease (%)" #?
     disease_percentage_evolution: "Disease percentage evolution"
     display: "Display" #?
+    display_n_1: "Display N-1"
```
Aquaj: See question on French locale.

app/views/backend/trial_balances/show.html.haml
```diff
@@ -31,44 +34,65 @@
...
         - elsif item[1].to_i == -1
           %tr.total
             %th{colspan: 2}= :total.tl
             %td.dec= number_to_accountancy(item[2], currency)
             %td.dec= number_to_accountancy(item[3], currency)
+            %td.dec= number_to_accountancy(prev_item[2], currency)
+            %td.dec= number_to_accountancy(prev_item[3], currency)
             %td.dec= number_to_accountancy(item[4], currency) if item[4].to_f > 0
             %td.dec= number_to_accountancy(-item[4].to_f, currency) if item[4].to_f < 0
+            %td.dec= number_to_accountancy(prev_item[4], currency) if prev_item[4].to_f > 0
```
Aquaj: No `to_f` here? What's the difference with the next line?

app/views/backend/trial_balances/show.html.haml
```diff
@@ -31,44 +34,65 @@
...
           %tr{class: cycle(:odd, :even)}
             %td.code= link_to(account.number, action: :show, controller: :accounts, id: account.id)
             %td= link_to(account.name, action: :show, controller: :accounts, id: account.id)
             %td.dec= number_to_accountancy(item[2], currency)
             %td.dec= number_to_accountancy(item[3], currency)
+            %td.dec= number_to_accountancy(prev_item[2], currency)
+            %td.dec= number_to_accountancy(prev_item[3], currency)
             %td.dec= number_to_accountancy(item[4], currency) if item[4].to_f > 0
             %td.dec= number_to_accountancy(-item[4].to_f, currency) if item[4].to_f < 0
+            %td.dec= number_to_accountancy(prev_item[4], currency) if prev_item[4].to_f > 0
```
Aquaj: No `to_f` here? What's the difference with the next line?

app/controllers/backend/trial_balances_controller.rb
```diff
@@ -26,81 +26,118 @@ def show
       document_nature = Nomen::DocumentNature.find(:trial_balance)
       key = "#{document_nature.name}-#{Time.zone.now.l(format: '%Y-%m-%d-%H:%M:%S')}"
       filename = document_nature.human_name
+      if params[:period] == 'all'
+        params[:started_on] = FinancialYear.all.map { |fy| fy.started_on }.sort.first.to_s
```
Aquaj: If you're going to map over all to get a column attribute, use `pluck` instead. You can even combine `order` and `pluck` to avoid the `sort` too.

app/controllers/backend/trial_balances_controller.rb
```diff
@@ -26,81 +26,118 @@ def show
...
+        params[:stopped_on] = FinancialYear.all.map { |fy| fy.stopped_on }.sort.last.to_s
```
Aquaj: If you're going to map over all to get a column attribute, use `pluck` instead. You can even combine `order` and `pluck` to avoid the `sort` too.

config/locales/fra/action.yml
Zank94: The idea is to display data for the same selected period one year before, which may not be the previous financial year, not to mention it can also be a custom interval where "previous period" would become inaccurate

app/views/backend/trial_balances/show.html.haml
Zank94: Good question, to be honest I just copy/pasted
`%td.dec= number_to_accountancy(item[4], currency) if item[4].to_f > 0`
`%td.dec= number_to_accountancy(-item[4].to_f, currency) if item[4].to_f < 0`

PR #2321, opened 2018-10-18
Debug/fix general ledger

DucMirack:


app/controllers/backend/base_controller.rb
```diff
@@ -361,11 +361,9 @@ def account_journal_period_crit(variable, conditions = 'c')
       def centralizing_account_crit(variable, conditions = 'c')
         variable = "params[:#{variable}]" unless variable.is_a? String
         code = ''
-        code << "if #{variable}[:ledger] == 'general_ledger'\n"
-        code << "  #{conditions}[0] += \" AND accounts.nature != 'auxiliary'\"\n"
-        code << "elsif centralizing_account = Account.find_by(number: #{variable}[:ledger])\n"
+        code << "if #{variable}[:ledger] != 'general_ledger' && auxiliary_accounts = Account.get_auxiliary_accounts(#{variable}[:ledger])\n"
         code << "  #{conditions}[0] += \" AND accounts.nature = 'auxiliary'\"\n"
-        code << "  #{conditions}[0] += ' AND accounts.centralizing_account_id = '+ centralizing_account.id.to_s\n"
+        code << "  #{conditions}[0] += \" AND accounts.id IN (\#{auxiliary_accounts.pluck(:id).join(', ')})\"\n"
```
Zank94: I think the `.join(', ')` is not necessary here

Aquaj: It is. If you don't you end up with `AND accounts.id IN ([1, 2, 3])` instead of `AND accounts.id IN (1, 2, 3)` and postgres doesn't liek it much.

PR #2322, opened 2018-10-18
Feature/accountancy 2018/fec export

Zank94:


app/jobs/fec_export_job.rb
```diff
@@ -0,0 +1,22 @@
+class FecExportJob < ActiveJob::Base
+  queue_as :default
+  include Rails.application.routes.url_helpers
+
+  def perform(financial_year, fiscal_position, interval, user)
```
DucMirack:
```suggestion
  def perform(financial_year, fiscal_position, interval = 'year', user)
```

app/jobs/fec_export_job.rb
```diff
@@ -0,0 +1,22 @@
...
+    interval ||= 'year'
```
DucMirack:
```suggestion
```

test/models/financial_year_test.rb
```diff
@@ -208,4 +208,33 @@ class FinancialYearTest < ActiveSupport::TestCase
       assert f.close(nil, options), "Financial year #{f.code} should be closed"
     end
   end
+
+  test 'compute periods given a specific interval' do
+    financial_year_18_19 = FinancialYear.find_by_code('EX2018/2019')
```
DucMirack: What if the given financial year doesn't exists? Shouldn't you create it first ?

Zank94: Well it comes from the fixtures (pre loaded)

PR #2327, opened 2018-10-23
Feature/accountancy 2018/two financials years opened

DucMirack:


app/models/financial_year.rb
```diff
@@ -276,7 +275,7 @@ def next
 
   # this method returns the previous financial year record sorted by started_on
   def previous_record
-    all_fy = self.class.all.sort_by(&:started_on)
+    all_fy = FinancialYear.order(:started_on)
     return nil if all_fy.index(self).zero?
     self_index = all_fy.index(self)
```
Aquaj: You go around the collection twice, not so gret for perfs. You should do something like:
```suggestion
self_index = all_fy.index(self)
return nil if self_index.zero?
previous_record = all_fy[self_index - 1]
```

PR #2328, opened 2018-10-24
Debug/payment delay

claire-gtr:


lib/delay.rb
```diff
@@ -6,37 +6,43 @@ class InvalidDelayExpression < ArgumentError
...
+      'hour' => :hour,
+      'hours' => :hour,
+      'minute' => :minute,
+      'minutes' => :minute,
+      'second' => :second,
+      'seconds' => :second,
+    }
   }.freeze
-  KEYS = TRANSLATIONS.keys.join('|').freeze
+  KEYS = TRANSLATIONS[:fra].merge(TRANSLATIONS[:eng]).keys.join('|').freeze
```
Aquaj: We should go for a key-agnostic approach here:
```suggestion
TRANSLATIONS.values.reduce(&:merge).keys.join('|').freeze
```
That way if we ever add another key it'll still work.

Also, since we use that "full translation hash" later on in the code, rather than re-merging each time I think we should extract it into a constant. Maybe `KEYWORDS` or `ALL_TRANSLATIONS` or something like that?

config/locales/eng/models.yml
```diff
@@ -190,6 +190,8 @@ eng:
           unvalid_amounts: "The journal entry must be filled either debit or credit but not at the same time."
         outgoing_parcel_item:
           greater_than_undelivered_quantity: "The quantity of %{product} to deliver is greater than the available quantity (%{maximum}%{unit})"
+        purchase:
+          payment_delay_custom_validation_message: "Correct formats : eom, end of month, %{values}"
```
Aquaj: `eom` and `end of month` should be in the interpolation too.

(With something like `values: TRANSLATIONS[...].... + ["eom", "end of month"]` probably)

Also, `bom`, `beginning of month` and `ago` should be in the list too.

config/locales/fra/models.yml
```diff
@@ -326,6 +326,8 @@ fra:
           unvalid_amounts: "L’écriture doit être renseignée soit au débit, soit au crédit mais pas les deux en même temps."
         outgoing_parcel_item:
           greater_than_undelivered_quantity: "La quantité de produit %{product} à livrer dépasse la quantité non livrée (%{maximum}%{unit})"
+        purchase:
+          payment_delay_custom_validation_message: "Formats acceptés : fdm, fin du mois, %{values}"
```
Aquaj: `fdm` and `fin de mois` should be in the interpolation too.

(With something like `values: TRANSLATIONS[...].... + ["fdm", "fin du mois"]` probably)

Also, `début du mois`, `ddm` and `avant` should be in the list too.

lib/delay.rb
```diff
@@ -151,7 +157,7 @@ module Validator
     module ClassMethods
       def validates_delay_format_of(*attr_names)
         attr_names.each do |attr_name|
-          validates attr_name, delay: true
+          validates attr_name, delay: { message: proc { I18n.translate('activerecord.errors.models.purchase.payment_delay_custom_validation_message', values: (TRANSLATIONS[Preference[:language].to_sym]&.keys.join(', ') || TRANSLATIONS[:eng].keys.join(', ') )) } }
```
Aquaj: This is way too long and hard to read.
I think a good idea could be either to set the proc into a constant in the class:
```suggestion
BAD_DELAY_MESSAGE_PROC = proc do 
  # ...
end

  # ...

  validates attr_name, delay: { message: BAD_DELAY_MESSAGE_PROC }
```
or into a class method:
```suggestion
  def self.bad_delay_message(*args)
     # ...
  end

  # ...
  validates attr_name, delay: { message: method(:bad_delay_message) }
```
Either way once extracted it should be refactored it into a multiline expression using properly named variables.

PR #2333, opened 2018-10-31
Feature/accountancy 2018/financial year close

DucMirack:
- On close view, add warnings and checks, if any check fails, it is impossible to close the financial year.
- There are 4 checks which verify conditions to enable the close : 
1. All previous years are all closed or locked
2. There is no draft entry on the current financial year period
3. There is no entry to balance on the current financial year period
4. Radical account classes (accounts with number beginning by 1 to 7 and with leading 0, for example: 1, 100, 2, 30, 4000000, 500, ...) have their journal_entry_items balanced on the current financial year period.

- If a check fails, there is a link to redirect to the page to resolve the given problem, if possible parameters like period or state are passed to filter elements.
- A warning doesn't block the close but enable to inform the user of the irreversible nature of the close operation.
- A warning appear if : 
1. There are draft fixed assets
2. There are draft loans
3. There are not sent tax declarations on the current financial year period
4. There are not given parcel (incoming and outgoing) on the current financial year period
5. There are not invoiced sale on the current financial year period
6. There are not invoiced purchase on the current financial year period

- Same as for the checks, if a warning appear, there is a link to redirect the user.
- A line presents the result for the current financial year
- As for the lock page, there is a checkbox to check in order to click the close button which himself asks to confirm the action.


app/models/account.rb
```diff
@@ -90,7 +90,7 @@ class Account < Ekylibre::Record::Base
   validates :last_letter, length: { allow_nil: true, maximum: 10 }
   validates :name, length: { allow_nil: true, maximum: 200 }
   validates :number, uniqueness: true
-  validates :number, length: { is: 8 }, format: { without: /\A0/ }, unless: :already_existing_and_general
+  validates :number, length: { is: Preference[:account_number_digits] }, format: { without: /\A0/ }, unless: :already_existing_and_general
```
Aquaj: This doesn't work. You need to put this inside a block or it won't get evaluated at validation time.

app/models/financial_year.rb
```diff
@@ -308,23 +309,9 @@ def sum_entry_items(expression, options = {})
...
-  end
-
   def sum_entry_items_with_mandatory_line(document = :profit_and_loss_statement, line = nil, options = {})
-    # remove closure entries
-    options[:unwanted_journal_nature] ||= [:closure] if document == :balance_sheet
-    options[:unwanted_journal_nature] ||= %i[result closure]
-
-    equation = get_mandatory_line_calculation(document, line) if line
-    equation ? sum_entry_items(equation, options) : 0
+    # deprecated use AccountancyComputation sum_entry_items_by_line instead
```
Aquaj: 👍

app/models/financial_year.rb
```diff
@@ -449,6 +436,64 @@ def split_into_periods(interval)
...
+        balanced_class[:period] = self.started_on.to_s + '_' + self.stopped_on.to_s
+        balanced_radical_account_classes << balanced_class
+      end
+    end
+    balanced_radical_account_classes
+  end
+
+  def any_invalid_closure_check?
+    checks = [all_previous_financial_years_closed_or_locked?, no_draft_entry?, no_entry_to_balance?, unbalanced_radical_account_classes_array.empty?]
+    checks.any? { |c| c == false } ? true : false
```
Aquaj: `(condition) ? true : false` == `condition`
(and `(condition) ? false : true` == `!condition`, basically anytime you write a ternary/if-else whose results are `true` and `false`, you're missing the opportunity to just use the condition as you result.)

app/services/financial_year_close.rb
```diff
@@ -11,16 +11,60 @@ def initialize(year, to_close_on, options = {})
...
+    # puts t
+    Rails.logger.info(t)
+    now
+  end
+
+  def log(message)
+    @previous_now = say(message)
+  end
+
+  def cinc(name, _increment = 1)
```
Aquaj: If we're done testing the FinancialYearClose, I'd like that we remove all the testing code that was set up by @burisu. Unless we couple it to a proper reporting system, it has no place in production.

app/services/financial_year_close.rb
```diff
@@ -11,16 +11,60 @@ def initialize(year, to_close_on, options = {})
     @options = options
   end
 
+  def benchmark(message)
```
Aquaj: Was the close properly tested with a big amount of data to confirm performance?

app/views/backend/financial_years/close.html.haml
```diff
@@ -1,11 +1,67 @@
 = form_tag do
-  = field_set do
+  = field_set :prerequisites_checks_before_closure.tl do
+    .close-conditions
+      -# ------ Warnings ------
```
Aquaj: Having a `Warnings` comment right abover a `warnings` helper call feels redundant.

app/views/backend/financial_years/close.html.haml
```diff
@@ -1,11 +1,67 @@
...
+      = warnings(@financial_year,
+                  fixed_assets: FixedAsset.drafts,
+                  loans: Loan.drafts,
+                  tax_declarations: :not_sent_tax_declarations,
+                  incoming_parcels: :not_given_incoming_parcel,
+                  purchases: :not_invoiced_or_aborted_purchase,
+                  outgoing_parcels: :not_given_outgoing_parcel,
+                  sales: :not_invoiced_or_aborted_sale)
+
+      -# ------ Checks ------
```
Aquaj: Having a `Checks` comment right above a `checks` helper call feels redundant.

config/environments/development.rb
```diff
@@ -40,13 +40,13 @@
...
   if defined? Bullet
-    config.after_initialize do
-      Bullet.enable = true
-      Bullet.alert = false
-      Bullet.bullet_logger = true
-      Bullet.console = true
-      Bullet.add_footer = true
-      Bullet.rails_logger = true
-    end
+    # config.after_initialize do
```
Aquaj: This shouldn't be in the PR.

app/services/financial_year_close.rb
DucMirack: I don't know, @ionosphere was in charge of testing the closure, should we wait for him to tell us if we remove it or not ?

PR #2345, opened 2018-11-27
Feature/accountancy 2018/add end first finyear firstrun

claire-gtr:


app/controllers/backend/financial_years_controller.rb
```diff
@@ -90,9 +90,9 @@ def show
     def new
       @financial_year = FinancialYear.new
       f = FinancialYear.last
-      @financial_year.started_on = f.stopped_on + 1 unless f.nil?
+      @financial_year.started_on = f.nil? ? Entity.of_company&.born_at : f.stopped_on + 1
       @financial_year.started_on ||= Time.zone.today
-      @financial_year.stopped_on = ((@financial_year.started_on - 1) >> 12).end_of_month
+      @financial_year.stopped_on = f.nil? && Entity.of_company&.first_financial_year_ends_on ? Entity.of_company.first_financial_year_ends_on : ((@financial_year.started_on - 1) >> 12).end_of_month
```
Aquaj: This is a tad complex. Can we use a `if`/`else` instead?
Also can we use Rails' `+ 1.year` syntax instead of `>> 12` ?

app/controllers/backend/financial_years_controller.rb
```diff
@@ -90,9 +90,13 @@ def show
     def new
       @financial_year = FinancialYear.new
       f = FinancialYear.last
-      @financial_year.started_on = f.stopped_on + 1 unless f.nil?
+      @financial_year.started_on = f.nil? ? Entity.of_company&.born_at : f.stopped_on + 1
       @financial_year.started_on ||= Time.zone.today
-      @financial_year.stopped_on = ((@financial_year.started_on - 1) >> 12).end_of_month
+      if f.nil? && Entity.of_company&.first_financial_year_ends_on
```
Aquaj: I don't get why we're checking on `f.nil?` here? It's not called in any of the clauses.

PR #2353, opened 2018-12-04
Validator and JS date picker changes for first run creation date

claire-gtr:


app/validators/date_format_validator.rb
```diff
@@ -0,0 +1,13 @@
+class DateFormatValidator < ActiveModel::EachValidator
+  def validate_each(record, attribute, value)
+    unless value.split(/\D/).size == 3
+        record.errors.add(attribute, :invalid)
```
Aquaj: You're one step too indented there.

app/assets/javascripts/form/dates.js.coffee
```diff
@@ -1,15 +1,31 @@
...
+      $(".text-datepicker").each ->
+        $(this).attr('type', 'text')
+
+    $(document).on "focusin click keyup change", "input[type='text'].text-datepicker", (event) ->
+      element = $(this)
+      if element.attr("autocomplete") isnt "off"
+        locale = element.attr("lang") or I18n.locale.slice(0, 2) # until we get corresponding locale codes
+        options = {}
+        $.extend options, $.datepicker.regional[locale],
+          dateFormat: "dd-mm-yy"
```
Aquaj: Can we put the `dateFormat` line before the `$.datepicker.regional[locale]` one? This should ensure dates are properly localized instead of always `dd-mm-yy`.
We will end up with `dd/mm/yy`though, so maybe we need to add something like
```suggestion
  $.datepicker.regional["fr"].dateFormat = "dd-mm-yy"
```
at the top of the file ?

PR #2369, opened 2018-12-12
Don't validate siret if company is not french

piotaixr:


app/models/entity.rb
```diff
@@ -175,7 +175,7 @@ class Entity < Ekylibre::Record::Base
   validates :activity_code, length: { allow_nil: true, maximum: 30 }
   validates :deliveries_conditions, :number, length: { allow_nil: true, maximum: 60 }
   validates :iban, iban: true, allow_blank: true
-  validates :siret_number, siret_format: true, allow_blank: true
+  validates :siret_number, siret_format: {if: country == 'fr'}, allow_blank: true
```
Aquaj:
```suggestion
   validates :siret_number, siret_format: { if: -> (record) { record.country == 'fr' } }, allow_blank: true 
```

PR #2371, opened 2018-12-13
Now preventing to create a new loan if the ongoing_at field is during…

Zank94:
… a closed financial year

### DIFF
--- a/config/locales/eng/models.yml
+++ b/config/locales/eng/models.yml
@@ -153,6 +153,11 @@ eng:
     errors:
       messages:
         <<: *error_messages
+      models:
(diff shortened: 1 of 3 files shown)


config/locales/eng/models.yml
```diff
@@ -153,6 +153,11 @@ eng:
     errors:
       messages:
         <<: *error_messages
+      models:
+        loan:
+          attributes:
+            ongoing_at:
+              not_opened_financial_year: "Un exercice comptable ouvert est nécessaire"
```
Aquaj: 🇬🇧 please.

PR #2374, opened 2018-12-14
Feature/dom mutation watcher

piotaixr:


app/assets/javascripts/ekylibre.js.coffee
```diff
@@ -36,6 +36,28 @@
...
+  listeners = []
+  # Wait for dom loaded event before starting to watch. Otherwise we get A LOT of events.
+  $ =>
+    observer = new MutationObserver (mutationList, observer) ->
+      for mutation in mutationList
+        for listener in listeners
+          $element = $(listener.selector, mutation.addedNodes)
+          listener.callback $element if $element.length
+      return
+    observer.observe(document, {childList: true, subtree: true})
```
Aquaj: So by default we have a constant, recursively-watching, without any element scoping observer running ?
I find this a bit risky, performance wise. Can we profile that on some of our big pages, especially some that are updated with ajax calls and stuff?

piotaixr: Changed: put everything in a class + watch only if there is a listener

PR #2392, opened 2018-12-20
Updating an affairable model entity_id also updates the corrrespondin…

Zank94:
…g affair third_id


app/models/affair.rb
```diff
@@ -154,6 +154,16 @@ def affairable_types
       @affairable_types ||= %w[SaleGap PurchaseGap Sale Purchase Payslip IncomingPayment OutgoingPayment Regularization DebtTransfer].freeze
     end
 
+    def affairable_types_entity_foreign_keys
```
Aquaj: We should forego this and use polymorphism instead (I think we already do with the name `third`.)

app/models/affair.rb
```diff
@@ -240,9 +250,19 @@ def status
...
   def refresh!
     reload
+    update_third
     save!
   end
 
+  def update_third
+    entities_ids = []
+    self.class.affairable_types_entity_foreign_keys.each do |model, foreign_key|
+      entities_ids << model.to_s.camelize.constantize.where(affair_id: id).pluck(foreign_key)
```
Aquaj: Polymorphism would make this something like `model.where(affair: self).pluck(:third_id)`
(Maybe we need to look into attribute aliasing, I'm pretty sure ARecord can do that.

app/models/affair.rb
```diff
@@ -154,6 +154,10 @@ def affairable_types
       @affairable_types ||= %w[SaleGap PurchaseGap Sale Purchase Payslip IncomingPayment OutgoingPayment Regularization DebtTransfer].freeze
     end
 
+    def affairable_types_with_third
```
Aquaj: Why not make this a bit more dynamic:
```ruby
@affairable_types.map(&:constantize).select { |model| model.respond_to? :third_id }
```

Aquaj: I even wonder if we do need the method at all since it's only used once. Maybe a variable instead? What do you think?

PR #2396, opened 2018-12-20
Importing a different list of ProductNatureVariant when setting up a farm that uses fr_pcg82

piotaixr:


app/models/product_nature_variant.rb
```diff
@@ -533,9 +533,13 @@ def import_from_nomenclature(reference_name, force = false)
...
+      variants_to_load = Nomen::ProductNatureVariant.all
+      variants_to_load = pcg82_variants if _options.fetch(:preferences, {}).fetch(:accounting_system, '') == 'fr_pcg82'
```
Aquaj: If `_options` is used it shouldn't be prefixed with an underscore.

lib/ekylibre/first_run/folder.rb
```diff
@@ -60,12 +60,12 @@ def load_preferences
...
 
       # Load default data of models with default data
-      def load_defaults
+      def load_defaults(**_options)
         default_datasets.each do |dataset|
           next if @defaults[dataset].is_a?(FalseClass)
           puts "Load default #{dataset}..."
           model = dataset.to_s.classify.constantize
-          model.load_defaults
+          model.load_defaults(**_options, preferences: @preferences)
```
Aquaj: If `_options` is used it shouldn't be prefixed with an underscore.

piotaixr:
force push: rebase on develop/accountancy-2018

PR #2413, opened 2019-02-06
Several improvements to general ledger export

Zank94:


app/controllers/backend/general_ledgers_controller.rb
```diff
@@ -75,7 +75,7 @@ def self.subledger_accounts_selections
...
+      q1 = %q{Account.select("rpad(accounts.number, 8, '0') AS account_number, accounts.centralizing_account_name AS account_name, CASE WHEN (SUM(journal_entry_items.real_debit) - SUM(journal_entry_items.real_credit)) >= 0 THEN SUM(journal_entry_items.real_debit) - SUM(journal_entry_items.real_credit) ELSE 0 END AS cumulated_absolute_debit_balance, CASE WHEN (SUM(journal_entry_items.real_debit) - SUM(journal_entry_items.real_credit)) < 0 THEN @ SUM(journal_entry_items.real_debit) - SUM(journal_entry_items.real_credit) ELSE 0 END AS cumulated_absolute_credit_balance, journal_entry_items.absolute_currency AS account_currency").joins('INNER JOIN "journal_entry_items" ON "journal_entry_items"."account_id" = accounts."id"').joins('INNER JOIN "journal_entries" ON "journal_entries"."id" = "journal_entry_items"."entry_id"').where("nature = 'auxiliary'").group('account_name, account_number, account_currency')}
```
Aquaj: What has changed here, and why?

Zank94: The way centralizing accounts are displayed within the list.
It used to be only the first three digits, and now they follow the 8 digits convention.

Aquaj: Alright, makes sense.

PR #2416, opened 2019-02-08
Debug/accountancy 2018/pdf exports

Zank94:


Aquaj:
What bug is this fixing?

PR #2424, opened 2019-02-25
[Exception] matters#create (NoMethodError) undefined method read_at f…

pquintrie:
…or nil:NilClass

### DIFF
--- a/app/models/product_reading.rb
+++ b/app/models/product_reading.rb
@@ -91,7 +91,7 @@ class ProductReading < Ekylibre::Record::Base
   end
 
   after_save do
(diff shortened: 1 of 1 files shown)


app/models/product_reading.rb
```diff
@@ -91,7 +91,7 @@ class ProductReading < Ekylibre::Record::Base
   end
 
   after_save do
-    last_reading = product.readings.where(indicator_name: indicator_name).last
+    last_reading = ProductReading.where(product_id: product.id, indicator_name: indicator_name).last
```
piotaixr: last_reading can still be Nil here (no ProductReading)

pquintrie: i let you see what i did to fix the issue

PR #2425, opened 2019-02-26
[Exception] land_parcels#index (ActionView::Template::Error) undefine…

pquintrie:
…d method shape_centroid for nil:NilClass

### DIFF
--- a/test/helpers/backend/land_parcels_helper_test.rb
+++ b/test/helpers/backend/land_parcels_helper_test.rb
@@ -2,31 +2,31 @@ require 'test_helper'
 
 module Backend
   class LandParcelsHelperTest < ActionView::TestCase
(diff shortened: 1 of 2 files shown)


test/helpers/backend/land_parcels_helper_test.rb
```diff
@@ -2,6 +2,32 @@
 
 module Backend
   class LandParcelsHelperTest < ActionView::TestCase
-    # Add tests here...
+    class Bidule
```
piotaixr: :smile: Maybe rename 'Bidule' (and following variables like _truc, etc.) ?

pquintrie: Did it, you can check it !

PR #2433, opened 2019-03-05
[Exception] sales#update (Ekylibre::Record::RecordNotUpdateable) Reco…

pquintrie:
…rd cannot be updated

### DIFF
--- a/app/controllers/backend/base_controller.rb
+++ b/app/controllers/backend/base_controller.rb
@@ -129,7 +129,12 @@ module Backend
     def save_and_redirect(record, options = {})
       record.attributes = options[:attributes] if options[:attributes]
       ActiveRecord::Base.transaction do
-        if options[:saved] || record.send(:save)
+        can_be_saved = if record.new_record?
+                         record.createable?
+                       else
+                         record.updateable?
+                       end
+        if can_be_saved && (options[:saved] || record.send(:save))
           response.headers['X-Return-Code'] = 'success'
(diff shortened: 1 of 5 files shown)


app/controllers/backend/base_controller.rb
```diff
@@ -129,7 +129,12 @@ def find_and_check(*args)
     def save_and_redirect(record, options = {})
       record.attributes = options[:attributes] if options[:attributes]
       ActiveRecord::Base.transaction do
-        if options[:saved] || record.send(:save)
+        can_be_saved = if record.new_record?
```
piotaixr: Can maybe be simplified by 
`can_be_saved = (record.new_record? && record.createable?) || record.updateable?`

I dunno if it's really more readable, though. @Aquaj @Zank94 a thought on this?

app/controllers/backend/base_controller.rb
```diff
@@ -129,7 +129,12 @@ def find_and_check(*args)
...
+                         record.createable?
+                       else
+                         record.updateable?
+                       end
+        if can_be_saved && (options[:saved] || record.send(:save))
```
piotaixr: As the line is modified... `record.save` ?

app/controllers/backend/base_controller.rb
Aquaj: I think a ternary operator would work here, the `if/else` does feel a bit clunky, but the `(... && ...) || ...` feels hard to parse if you're not used to creative applications of lazy evaluation.

PR #2438, opened 2019-03-11
Debug/ruby 2.6 tests monkeypatch+fixes

piotaixr:


app/concepts/readings_coder.rb
```diff
@@ -4,7 +4,7 @@ class ReadingsCoder
                     .tap { |h| h[Charta::Point]        = proc { |_, value| value.to_ewkt } }
                     .tap { |h| h[Charta::MultiPolygon] = proc { |_, value| value.to_ewkt } }
 
-  UNSERIALIZE = Hash.new { proc { |klass, value| klass.new(value) } }
+  UNSERIALIZE = Hash.new { proc { |klass, value| klass == BigDecimal ? BigDecimal(value) : klass.new(value) } }
```
Aquaj: This should be made by adding a `.tap { |h| h[BigDecimal] = proc { ... } }` line and keep with the pattern instead of bloating the default proc.

piotaixr: Done

PR #2440, opened 2019-03-11
Change 'Accountable' column in fixed_assets_depreciations list to 'Accounted'

piotaixr:


app/models/fixed_asset_depreciation.rb
```diff
@@ -95,6 +95,10 @@ class FixedAssetDepreciation < Ekylibre::Record::Base
     end
   end
 
+  def accounted
+    !locked && accountable
```
Aquaj: A FixedAssetDepreciation is considered "accounted" when it isn't locked and when it is accountable ?
Shouldn't something that's been "accounted" **not** be accountable anymore?

piotaixr: If i understand the code correctly, once it has been flagged as 'accountable', it is never set back to *not* 'accountable'

PR #2442, opened 2019-03-12
Fixed FEC export

Zank94:


app/concepts/fec/exporter/xml.rb
```diff
@@ -60,7 +60,7 @@ def build(journals)
             end
           end
         end
-        builder.to_xml
+        builder.to_xml.gsub("&#8217;", "'")
```
piotaixr: https://www.rubydoc.info/github/sparklemotion/nokogiri/Nokogiri/XML/Node#to_xml-instance_method
You should set the encoding in the to_xml method :-) U+8217 is the Unicode code for the `'` character.

Zank94: `builder.to_xml(encoding: 'ISO-8859-15')`
doesn't work either, unfortunately

app/jobs/fec_export_job.rb
```diff
@@ -11,7 +11,11 @@ def perform(financial_year, fiscal_position, interval, user)
         fec = FEC::Exporter::XML.new(financial_year, fiscal_position, period.first, period.last)
         file_path = Ekylibre::Tenant.private_directory.join('tmp', "#{filename}")
         FileUtils.mkdir_p(file_path.dirname)
-        File.write(file_path, fec.generate)
+
+        File.open(file_path, "wb:ISO-8859-15") do |fout|
```
Aquaj: 😮

Aquaj: (I didn't know you could do that)

PR #2443, opened 2019-03-12
Added conditions to loans and repayments edition

Zank94:


app/controllers/backend/loans_controller.rb
```diff
@@ -55,7 +55,7 @@ def self.list_conditions
     end
 
     list(conditions: list_conditions, selectable: true) do |t|
-      t.action :edit, if: :updateable?
+      t.action :edit, if: :updateable? && :draft?
```
Aquaj: This won't work. `:updateable? && :draft?` always return `:draft` (`truthy && truthy`)
You either need to make a method out of those and call that, or to use a `CodeString` (`"RECORD.updateable? && RECORD.draft?".c`)

PR #2452, opened 2019-03-13
Fix bug preventing to create a FixedAsset using depreciation_method: :none

piotaixr:


app/models/fixed_asset.rb
```diff
@@ -89,22 +89,23 @@ class FixedAsset < Ekylibre::Record::Base
...
+  validates :stopped_on, timeliness: { on_or_after: ->(fixed_asset) { fixed_asset.started_on || Time.new(1, 1, 1).in_time_zone }, on_or_before: -> { Time.zone.today + 50.years }, type: :date }, allow_blank: true
   # ]VALIDATORS]
   validates :name, uniqueness: true
   validates :depreciation_method, inclusion: { in: depreciation_method.values }
-  validates :asset_account, :expenses_account, presence: true
+  validates :asset_account, presence: true
   validates :currency, match: { with: :journal, to_invalidate: :journal }
   validates :depreciation_fiscal_coefficient, presence: true, if: -> { depreciation_method_regressive? }
   validates :started_on, financial_year_writeable: true, allow_blank: true
+  validates :stopped_on, :allocation_account, :expenses_account, presence: { unless: Proc.new { |fa| fa.depreciation_method.to_s == 'none' } }
```
Aquaj: Could we use an arrow lambda syntax here ?

app/models/fixed_asset.rb
```diff
@@ -375,26 +376,30 @@ def self.depreciate(options = {})
...
-    when /monthly/
-      starts += new_months
-    when /quarterly/
-      new_trimesters = new_months.select { |date| date.month.multiple_of? 3 }
-      starts += new_trimesters
-    when /yearly/
-      new_years = new_months.select { |date| date.month == 1 }
-      starts += new_years
+
+    unless depreciation_method.to_s == 'none'
```
Aquaj: Can we add `predicates: { prefix: true }` to the enumerize and then replace this with `depreciation_method_none?` ?

app/models/fixed_asset_depreciation.rb
```diff
@@ -53,22 +53,25 @@ class FixedAssetDepreciation < Ekylibre::Record::Base
...
 
   sums :fixed_asset, :depreciations, amount: :depreciated_amount
 
   bookkeep do |b|
     if fixed_asset.in_use?
-      b.journal_entry(fixed_asset.journal, printed_on: stopped_on.end_of_month, if: accountable && !locked) do |entry|
-        name = tc(:bookkeep, resource: FixedAsset.model_name.human, number: fixed_asset.number, name: fixed_asset.name, position: position, total: fixed_asset.depreciations.count)
-        entry.add_debit(name, fixed_asset.expenses_account, amount)
-        entry.add_credit(name, fixed_asset.allocation_account, amount)
+      unless fixed_asset.depreciation_method.to_s == 'none'
```
Aquaj: See the comment on `fixed_asset.rb`.

db/migrate/20190313140443_relax_fixed_asset_constraints_for_none_deprecation_method.rb
```diff
@@ -0,0 +1,15 @@
+class RelaxFixedAssetConstraintsForNoneDeprecationMethod < ActiveRecord::Migration
+  def change
+    reversible do |r|
+      r.up do
+        change_column :fixed_assets, :stopped_on, :date, null: true
```
Aquaj: Shouldn't this be done for `fixed_asset_depreciations` too?

db/structure.sql
```diff
@@ -1135,6 +1135,18 @@ CREATE SEQUENCE public.analysis_items_id_seq
 ALTER SEQUENCE public.analysis_items_id_seq OWNED BY public.analysis_items.id;
 
 
+--
+-- Name: ar_internal_metadata; Type: TABLE; Schema: public; Owner: -
+--
+
+CREATE TABLE public.ar_internal_metadata (
```
Aquaj: This is from Rails 5, how come we get it here ?

db/structure.sql
```diff
@@ -5676,6 +5688,43 @@ CREATE SEQUENCE public.product_nature_variant_readings_id_seq
 ALTER SEQUENCE public.product_nature_variant_readings_id_seq OWNED BY public.product_nature_variant_readings.id;
 
 
+--
+-- Name: product_nature_variant_valuings; Type: TABLE; Schema: public; Owner: -
+--
+
+CREATE TABLE public.product_nature_variant_valuings (
```
Aquaj: ?

db/migrate/20190313140443_relax_fixed_asset_constraints_for_none_deprecation_method.rb
piotaixr: Nope, because we don't generate any FixedAssetDepreciation when creating a FixedAsset that does not devaluate

Aquaj: Then you cannot relax the validation in the `FixedAssetDepreciation` model, since FixedAssetDepreciation#stopped_on still has the DB constraint `NOT NULL`.

app/models/fixed_asset.rb
```diff
@@ -105,7 +105,7 @@ class FixedAsset < Ekylibre::Record::Base
   validates :currency, match: { with: :journal, to_invalidate: :journal }
   validates :depreciation_fiscal_coefficient, presence: true, if: -> { depreciation_method_regressive? }
   validates :started_on, financial_year_writeable: true, allow_blank: true
-  validates :stopped_on, :allocation_account, :expenses_account, presence: { unless: Proc.new { |fa| fa.depreciation_method.to_s == 'none' } }
+  validates :stopped_on, :allocation_account, :expenses_account, presence: { unless: ->(fa) { fa.depreciation_method_none? } }
```
Aquaj: This can be replaced by `presence: { unless: :depreciation_method_none? }`.

Aquaj:
I resolved the conflicts but I'd feel better if you could recheck @piotaixr.

piotaixr:
If using `if depreciation_method_linear? || depreciation_method_regressive?` instead of `unless depreciation_method_none?` is OK for you, it is OK for me!

Aquaj:
Yeah it's fine by me, either makes sense in the context and conveys the intention.

PR #2453, opened 2019-03-14
[Exception] bank_statements#create (ArgumentError) comparison of Date…

pquintrie:
… with nil failed

### DIFF
--- a/app/models/bank_statement_item.rb
+++ b/app/models/bank_statement_item.rb
@@ -53,10 +53,11 @@ class BankStatementItem < Ekylibre::Record::Base
   validates :initiated_on, timeliness: { on_or_after: -> { Time.new(1, 1, 1).in_time_zone }, on_or_before: -> { Time.zone.today + 50.years }, type: :date }, allow_blank: true
   validates :letter, :memo, :transaction_number, length: { maximum: 500 }, allow_blank: true
   validates :name, presence: true, length: { maximum: 500 }
-  validates :started_on, presence: true
-  validates :stopped_on, presence: true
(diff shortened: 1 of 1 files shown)


app/models/bank_statement_item.rb
```diff
@@ -53,6 +53,8 @@ class BankStatementItem < Ekylibre::Record::Base
   validates :initiated_on, timeliness: { on_or_after: -> { Time.new(1, 1, 1).in_time_zone }, on_or_before: -> { Time.zone.today + 50.years }, type: :date }, allow_blank: true
   validates :letter, :memo, :transaction_number, length: { maximum: 500 }, allow_blank: true
   validates :name, presence: true, length: { maximum: 500 }
+  validates :started_on, presence: true
```
Aquaj: Do not add validations between `[VALIDATORS[ .... ]VALIDATORS]` (they are auto-generated). Add them before or after.

app/models/bank_statement_item.rb
```diff
@@ -71,7 +73,7 @@ class BankStatementItem < Ekylibre::Record::Base
     if (debit.nonzero? && credit.nonzero?) || (debit.zero? && credit.zero?)
       errors.add(:credit, :unvalid_amounts)
     end
-    if bank_statement && transfered_on
+    if bank_statement && transfered_on && started_on && stopped_on
```
Aquaj: This doesn't do what we want.
Ex:
`started_on`: nil
`stopped_on`: 14/03/2019
`transfered_on`: 15/03/2019
`transfered_on` > `stopped_on` so we should have an error.

Aquaj: Okay, started_on cannot be blank, alright.

PR #2458, opened 2019-03-14
Develop/tools export

Sanchodelniglo:


app/controllers/backend/exports_controller.rb
```diff
@@ -18,11 +18,21 @@
 
 module Backend
   class ExportsController < Backend::BaseController
+    include Backend::ExportsHelper
```
piotaixr: If the helper in in the standard helper directory, you can access its methods with `helpers.<method>' in the controllers, no need to include it.

Sanchodelniglo: done!!

PR #2459, opened 2019-03-15
Adding a Suspense account (471) in nomen

piotaixr:


db/structure.sql
```diff
@@ -1135,6 +1135,18 @@ CREATE SEQUENCE public.analysis_items_id_seq
 ALTER SEQUENCE public.analysis_items_id_seq OWNED BY public.analysis_items.id;
 
 
+--
```
Aquaj: Since only Nomen was changed, there's no reason for the `db/structure.sql` to have been modified.

PR #2462, opened 2019-03-19
Refactor fixed_asset_depreciation period computation with tests

piotaixr:


piotaixr:
the service DepreciacionCalculator was written to be able to handle the computation of the regressive and linear depreciation amounts but I haven't refactored this yet.
This can be done in this PR or in a separate one. @aquaj your thoughts on this?

test/services/depreciation_calculator_test.rb
```diff
@@ -0,0 +1,56 @@
+require 'test_helper'
+
+class FinancialYearCloseTest < ActiveSupport::TestCase
```
Aquaj: The class doesn't match the filename here.

app/services/depreciation_calculator.rb
```diff
@@ -0,0 +1,59 @@
...
+               return nil
+             end
+
+    groups.map { |_, v| ([v.first.first, v.last.second, v.map(&:third).reduce(&:+)]) }
+  end
+
+  def monthly_periods(started_on, yearly_percentage)
+    return if(yearly_percentage == 0)
+
+    depreciation_duration = 100 / yearly_percentage * 360 # in days
```
Aquaj: This will be automatically rounded if `yearly_percentage` happens to have been passed as an integer. Since we call it form the model for now it's not an issue but I'd feel better if we made it explicitly Float with a `to_f`.

app/services/depreciation_calculator.rb
```diff
@@ -0,0 +1,59 @@
...
+                 shifted = m.started_on - fy_shift
+                 [shifted.year, shifted.month % 3]
+               end
+             when :yearly
+               months.group_by { |m_started_on, _, _| (m_started_on - fy_shift).year }
+             else
+               return nil
+             end
+
+    groups.map { |_, v| ([v.first.first, v.last.second, v.map(&:third).reduce(&:+)]) }
```
Aquaj: This would be clearer with a multiline block and block-level variables. What `v` is, `v.first.first`, `v.last.second` and `v.map(&:third)` are isn't clear.

test/services/depreciation_calculator_test.rb
piotaixr: oops

PR #2465, opened 2019-03-20
Added some tasks to get job infos from sidekiq

Aquaj:


lib/tasks/job.rake
```diff
@@ -10,6 +10,73 @@ namespace :job do
...
   end
 
+  task running: :environment do
+    next false if Sidekiq::Stats.new.enqueued.zero?
+    job_class = ENV['JOB_CLASS']
+    job_tenant = ENV['TENANT']
+    running = false
+    Sidekiq::Queue.all.each do |queue|
+      jobs = queue.entries.select do |entry|
+        (!job_class || entry.item['args'].first['job_class'] == job_class) &&
```
piotaixr: Maybe refactor this in a lambda/proc/other?

Aquaj: Hard to do in the context of task definitions. (especially since they depend on the ENV variables)
We could do with a lambda-generating lambda, but I feel like this would add too much complexity for the benefits.

PR #2469, opened 2019-03-21
Debug/permanent stock inventory

Sanchodelniglo:


app/models/intervention.rb
```diff
@@ -386,6 +392,10 @@ def printed_on
     printed_at.to_date
   end
 
+  def during_financial_year?
+    FinancialYear.opened.where('? BETWEEN started_on AND stopped_on', printed_at).any?
```
piotaixr: Is Between inclusive? If yes, i think its a problem for stopped_on.

app/models/intervention_working_period.rb
```diff
@@ -152,6 +152,10 @@ def duration_gap
     (stopped_at - started_at) / 3600
   end
 
+  def during_financial_year_exchange?
+    FinancialYearExchange.opened.where('? BETWEEN started_on AND stopped_on', started_at).any?
```
piotaixr: Same here for Between

app/views/backend/interventions/_working_period_fields.html.haml
```diff
@@ -1,4 +1,4 @@
 - prefix = namespace + "[#{f.index}]"
 .nested-fields.working-period
   = link_to_remove_nested_association('working-period', f)
-  = f.datetime_range start_input_html: { class: "intervention-started-at", data: { intervention_updater: prefix + 'started_at' }}, stop_input_html: { data: { intervention_updater: prefix + 'stopped_at' }}
```
piotaixr: No need to remove this poor blank line ;-)

config/locales/fra/models.yml
```diff
@@ -460,6 +460,14 @@ fra:
           attributes: #?
             parameters: #?
               check_errored: "La connexion au partenaire a échoué. Veuillez vérifier les informations saisies ou réessayer plus tard." #?
+        intervention:
+          attributes:
+            printed_on:
+              not_opened_financial_year: "Un exercice comptable ouvert est nécessaire pour la (les) période(s) ci-dessous"
```
piotaixr: That's good to fix the locale problem for this particular model. But if you put the locale in `activemodel.errors` it is used for all models in one go!

app/models/intervention_working_period.rb
Aquaj: `BETWEEN` is inclusive on both end.

app/models/intervention.rb
Aquaj: `BETWEEN` is inclusive on both end.

PR #2471, opened 2019-03-22
Allow creation (import) of a FixedAsset on a year without an opened FinancialYear

piotaixr:


Aquaj:
@piotaixr Could you fix the conflicts (mainly the `bookkeep` one) ?

PR #2473, opened 2019-03-25
Added un-editable payment delay field in sales form

Zank94:


app/controllers/backend/sales_controller.rb
```diff
@@ -177,7 +177,7 @@ def new
       @sale = if params[:intervention_ids]
                 Intervention.convert_to_sale(params[:intervention_ids])
               else
-                Sale.new(nature: nature)
+                Sale.new(nature: nature, payment_delay: nature.payment_delay)
```
Aquaj: This should be applied to the `Intervention.convert_to_sale` too.
Could you maybe extract that to like:
```
@sale = ....
@sale.payment_delay = nature.payment_delay
``` ?

PR #2477, opened 2019-03-26
Made ManageRestfully generate modules dynamically

Aquaj:
This makes it so `manage_restfully`-generated methods can be overwritten and called through `super` instead of simply being overwritten when redefining:

Example:
```ruby
# Current behaviour
class MyController < ApplicationController
  manage_restfully only: :index

  def index
    Rails.log "Entering the index override"
    super # => NoMethodError: super: no superclass method `index' for `#<MyController:0x00007fbc5797fc78>
   end
end

# On PR
# Current behaviour
class MyController < ApplicationController
  manage_restfully only: :index

  def index
    Rails.log "Entering the index override"
    super # => returns index view
  end
end
```

### DIFF
--- a/app/controllers/concerns/restfully_manageable.rb
+++ b/app/controllers/concerns/restfully_manageable.rb
@@ -251,6 +251,7 @@ module RestfullyManageable
       end
 
       restful_module = Module.new
+      name_module(restful_module, '')
       restful_module.class_eval(code)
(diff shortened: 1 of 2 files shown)


Aquaj:
Yeah of course! Thanks for catching that @piotaixr, it's done.

PR #2489, opened 2019-04-01
Fixed no_VAT_declaration banner behavior

Sanchodelniglo:


app/controllers/backend/tax_declarations_controller.rb
```diff
@@ -64,7 +64,12 @@ def index
 
       respond_to do |format|
         format.html do
-          render "alert_no_VAT_declaration" unless FinancialYear.where.not(tax_declaration_mode: "none").any?
+          not_opened_financial_year = FinancialYear.opened.empty?
+          opened_financial_year_with_tax_declaration_mode_none = !FinancialYear.with_tax_declaration.include?(FinancialYear.current)
+          vat_declarations_fullfilled = !FinancialYear.current.missing_tax_declaration? if FinancialYear.current
+          render "alert_no_VAT_declaration" if not_opened_financial_year || opened_financial_year_with_tax_declaration_mode_none || vat_declarations_fullfilled
+          byebug
+          # FinancialYear.where.not(tax_declaration_mode: "none").any? || !FinancialYear.current.missing_tax_declaration?
```
Aquaj: Can we remove this comment?

app/controllers/backend/tax_declarations_controller.rb
```diff
@@ -64,7 +64,12 @@ def index
 
       respond_to do |format|
         format.html do
-          render "alert_no_VAT_declaration" unless FinancialYear.where.not(tax_declaration_mode: "none").any?
+          not_opened_financial_year = FinancialYear.opened.empty?
```
Aquaj: Can we rename this into something a bit more english? `no_financial_year_open` maybe?

app/controllers/backend/tax_declarations_controller.rb
```diff
@@ -64,7 +64,12 @@ def index
...
+          opened_financial_year_with_tax_declaration_mode_none = !FinancialYear.with_tax_declaration.include?(FinancialYear.current)
```
Aquaj: The variable name doesn't match its content. 
Also can we remove the `!` at the beginning and use the opposite predicate instead (`exclude?`)

app/controllers/backend/tax_declarations_controller.rb
```diff
@@ -64,7 +64,12 @@ def index
...
+          vat_declarations_fullfilled = !FinancialYear.current.missing_tax_declaration? if FinancialYear.current
```
Aquaj: If you don't have a `FinancialYear.current` this will make the `vat_declarations_fullfilled` variable `nil` (which is falsey) is this the intended behaviour?
If it is, I'd rather we make it explicit: `FinancialYear.current && !FinancialYear.current.missing_tax_declaration?`

app/controllers/backend/tax_declarations_controller.rb
```diff
@@ -64,7 +64,10 @@ def index
 
       respond_to do |format|
         format.html do
-          render "alert_no_VAT_declaration" unless FinancialYear.where.not(tax_declaration_mode: "none").any?
+          no_financial_year_opened = FinancialYear.opened.empty?
+          current_financial_year_with_tax_declaration = FinancialYear.with_tax_declaration.exclude?(FinancialYear.current)
```
Aquaj: This implies that the variable equals `true` when the current financial year has a TaxDeclaration, but you're actually looking for the opposite. Is this properly named?

app/controllers/backend/tax_declarations_controller.rb
```diff
@@ -64,7 +64,10 @@ def index
...
+          vat_declarations_fullfilled = FinancialYear.current && !FinancialYear.current.missing_tax_declaration?
+          render "alert_no_VAT_declaration" if not_opened_financial_year || opened_financial_year_with_tax_declaration_mode_none || vat_declarations_fullfilled
```
Aquaj: The variable names don't match with the previously assigned variables

Aquaj:
[Ouvrir /backend/general-ledger sans exercice comptable donne une page rouge.](https://trello.com/c/q15rWfeJ/31-ouvrir-backend-general-ledger-sans-exercice-comptable-donne-une-page-rouge)

PR #2516, opened 2019-04-23
Add accounts translations

Sanchodelniglo:


app/views/backend/accounts/show.html.haml
```diff
@@ -17,7 +17,7 @@
   - c.cobble(:entities) do
     = cobble_list(:entities)
 
-  - c.cobble(:product_nature_variants) do
+  - c.cobble(:product_nature_variant) do
```
Aquaj: Why has this been singularized? Plural seems to be what we want here.

PR #2523, opened 2019-04-29
Refactoring of state-bar component

piotaixr:


app/services/state_bar_builder.rb
```diff
@@ -0,0 +1,84 @@
...
+    @resource = resource
+    @attribute = attribute
+    @options = options
+  end
+
+  def build
+    transitions = extract_possible_transitions_from resource
+
+    buttons = states_from(resource)
+                .map { |state|
```
Aquaj: Multiline `map`s should use `do` syntax.

app/services/state_bar_builder.rb
```diff
@@ -0,0 +1,84 @@
...
+        .map { |c| transitions_mod.const_get c }
+        .select { |c| c < Transitionable::Transition }
+        .select { |t| t.from.include? resource.send(attribute).to_sym }
+        .map { |t| [t.to, t] }
+        .to_h
+    end
+
+    def extract_possible_transitions_from_state_machine(resource)
+      resource.state_transitions
+        .map { |t|
```
Aquaj: This can fit on a single-line.

app/helpers/state_bar_helper.rb
```diff
@@ -0,0 +1,91 @@
+module StateBarHelper
+  Colors = {
```
Aquaj: Could you upcase that, since it's not a class?

app/helpers/state_bar_helper.rb
```diff
@@ -0,0 +1,91 @@
...
+    intermediate: %i[draft],
+    rejected: %i[aborted lost refused scrapped sold],
+    validated: %i[done finished given in_use invoice ongoing order repaid won]
+  }.flat_map { |key, values| values.map { |v| [v, key] } }
```
Aquaj: Could this be moved to a method of the `Hash` class? (`reverse`? `flip`? Maybe `transpose`, similarly to `Array#transpose`)

app/services/state_bar_builder.rb
```diff
@@ -0,0 +1,84 @@
...
+                    current: (state.to_sym == resource.send(attribute).to_sym),
+                    transition: transition,
+                    label: rename(state)
+                  )
+                }
+
+    StateBarHelper::StateBar.new *buttons, transitions_enabled: transitions_enabled?
+  end
+
+
```
Aquaj: Can we remove some of those empty lines?

app/helpers/state_bar_helper.rb
piotaixr: I don't think moving this to the Hash class is a good idea as the implementation is really specific: what about if values are not unique?
What we will get if we add sth like `Hash#transpose` is a `MultiMap` and not a `Hash`

Aquaj: The target of this would be a category of `Hash`. The implementation would raise if there were non-unique values.
Ex in the Ruby Standard Lib: `Array#transpose` raises if the element array are not all of the same length.
```ruby
irb(main):002:0* [[:a, :b, :c], [:a, :b]].transpose
IndexError: element size differs (2 should be 3)
        from (irb):2:in `transpose'
```

piotaixr: heh, ok

PR #2531, opened 2019-05-02
Refactoring of test_helper

piotaixr:


lib/ekylibre/testing/application_controller_test_case.rb
```diff
@@ -0,0 +1,14 @@
+module Ekylibre
+  module Testing
+    class ApplicationControllerTestCase < ActionController::TestCase
+      include FactoryBot::Syntax::Methods
+      include Ekylibre::Testing::Concerns::FixturesCleanerModule
+
+      class WithFixtures < ActionController::TestCase
+        include FactoryBot::Syntax::Methods
```
Aquaj: Can we extract this into a module instead of a class?
(and then `include WithFixtures` in the test cases instead of having a very long class name to inherit from)

piotaixr:
squashed.

PR #2536, opened 2019-05-07
Develop/accounts translations

Sanchodelniglo:


config/locales/fra/action.yml
```diff
@@ -2116,6 +2116,7 @@ fra:
     product_category: "Produits"
     product_name: "Produit"
     product_natures: "Types de produit"
+    product_nature_variants: "Variante"
```
Aquaj: The key here is plural. The translation should be too.

Aquaj: Also: was it only missing in french? If it isn't present in the English locale it should be added there too.

config/locales/fra/action.yml
```diff
@@ -2116,6 +2116,7 @@ fra:
     product_category: "Produits"
     product_name: "Produit"
     product_natures: "Types de produit"
+    product_nature_variants: "Variantes"
```
piotaixr: I would put it before `product_natures`. (`_` < `s`)

PR #2539, opened 2019-05-09
Adds better error reporting to transitions

piotaixr:


test/models/concerns/transitionable_test.rb
```diff
@@ -135,6 +136,68 @@ def transition.transition
...
+        true
+      end
+
+      def t4.transition
+        true
+      end
+
+      assert t4.run
+      assert_nil t4.error
+
```
Aquaj: Can those empty lines be removed?

piotaixr: Done

PR #2551, opened 2019-06-07
Feature/fixed assets imports

Sanchodelniglo:

PR #2559, opened 2019-06-17
Feature gain and loss fixed assets exports

Sanchodelniglo:


app/exchangers/isagri/isacompta/fixed_assets_exchanger.rb
```diff
@@ -178,6 +101,31 @@ def import
           w.check_point
         end
       end
+
+      private
+
+      # Trim account number following preferences
+      def normalize(number)
+        preference_number = Preference[:account_number_digits]
+        differential = number.size - preference_number if number.size > preference_number
```
piotaixr: The conditions with `differential` are quite messy, can you reorganise them?

app/exchangers/isagri/isacompta/fixed_assets_exchanger.rb
```diff
@@ -178,6 +101,31 @@ def import
...
+          number
+        end
+      end
+
+      # Generate allocation account number
+      def to_allocation_account(number)
+        if number.first == "2"
+          number.chars.insert(1, "8")[0...-1].join
+        else
+          number
```
piotaixr: Is there a reason why the function defaults to the account number and not just fails when we cant 'convert' automatically to an account number?

app/exchangers/isagri/isacompta/fixed_assets_exchanger.rb
```diff
@@ -178,6 +101,31 @@ def import
           w.check_point
         end
       end
+
+      private
+
+      # Trim account number following preferences
+      def normalize(number)
```
piotaixr: Maybe also handle the case where the account number length is too short and extend it?
Then, it can be used in the initializer of the `Account` class.

app/models/fixed_asset.rb
```diff
@@ -261,14 +261,15 @@ def started_during_financial_year_closure_preparation?
...
   def self.depreciate(options = {})
-    depreciations = FixedAssetDepreciation.with_active_asset
+    depreciations = FixedAssetDepreciation.with_active_asset.not_locked.not_accountable
     depreciations = depreciations.up_to(options[:until]) if options[:until]
     transaction do
       # trusting the bookkeep to take care of the accounting
+      count = []
       depreciations.find_each do |dep|
-        dep.update!(accountable: true)
+        count << dep.update!(accountable: true)
```
piotaixr: You can just increment a variable here ;-)

db/migrate/20190606134257_change_usages_of_old_accounts_to_match_nomenclature.rb
```diff
@@ -0,0 +1,26 @@
...
+              fixed_assets: :financial_assets,
+              fixed_assets_suppliers: :financial_assets_suppliers ,
+              fixed_assets_values: :financial_assets_values }
+              
+  def up
+    CHANGED.each do |to, from|
+      execute <<-SQL
+        UPDATE accounts
+           SET usages = '#{to}'
+         WHERE usages = '#{from}'
```
piotaixr: the `usages` name seems to indicate that its possible to have multiple usages. @Aquaj ?

test/models/fixed_asset/fixed_asset_test.rb
```diff
@@ -127,8 +127,8 @@ class FixedAssetTest < Ekylibre::Testing::ApplicationTestCase
 
     test 'depreciate class method returns the amount of depreciations according to until option provided' do
       fixed_asset = create :fixed_asset, :yearly, :in_use, percentage: 100.0 / 3, started_on: Date.new(2017, 1, 1)
-
       count = FixedAsset.depreciate(until: Date.civil(2018, 12, 31))
+
```
piotaixr: :-) Not really necessary to commit this file

app/exchangers/isagri/isacompta/fixed_assets_exchanger.rb
piotaixr: Maybe add a test?

app/exchangers/isagri/isacompta/fixed_assets_exchanger.rb
```diff
@@ -178,6 +101,31 @@ def import
...
+          number.first(preference_number)
+        elsif differential && number.last != "0" * (differential)
+          number.first(6) + number.last(preference_number - 6)
+        else
+          number
+        end
+      end
+
+      # Generate allocation account number
+      def to_allocation_account(number)
```
piotaixr: A test for this method?

app/exchangers/isagri/isacompta/fixed_assets_exchanger.rb
Sanchodelniglo: It's seems to be already done in the before_validation of `Account`, that's why i just handled the "too long" case

piotaixr:
Depends on #2551

